PR #500, opened 2012-01-22
Add support for loading data from the EMI Mac installer

clone2727:
The EMI Mac installer has a StuffIt archive in its data fork, which contains four important files for the game: patch.m4b, Secrets.m4b, local.m4b, and AMWI.m4b. These two commits allow for loading the data from the installer archive so one does not need to install it beforehand.


andrews05:
What is the Secrets.m4b file? I have the Mac version and it's always puzzled me - it doesn't seem to contain any secrets, the file is only 8 bytes.

PR #719, opened 2013-02-09
GRIM: Don't save and restore last_set anymore

YakBizzarro:
With this commit ResidualVM doesn't save and restore the last_set key anymore, so even if developer mode is set, it start from the intro every time and not from the last set.

Does anyone find this feature useful and prefer the standard behavior?


aquadran:
implemented way to not store if it was not stored before

PR #721, opened 2013-03-23
This will not merge: GRIM: Joystick/gamepad support

inguin:
This commit adds true joystick/gamepad support to the Grim engine. Implementation of the new event types is currently available for the SDL backend only. The existing code which mapped joystick events to keyboard or mouse events is still present but disabled with a compile-time define.

Tested with X-Box 360 controller in Linux. Controls are slightly erratic when flipping through the inventory or when pushing the wheelbarrow, but otherwise it's a significant improvement over keyboard control.


chkr-private:
This pull request is obsoleted by #1147.

PR #752, opened 2013-08-25
EMI: Add the md5 of MonkeyUpdate.exe and check it, if present

YakBizzarro:
Adds the md5 of MonkeyUpdate.exe (and localized versions of it) and check it, instead of patch.m4b.
It also adds the md5s of i9n.m4b extracted from localized versions of update.

### DIFF
--- a/engines/grim/emi/emi.cpp
+++ b/engines/grim/emi/emi.cpp
@@ -48,6 +48,35 @@ LuaBase *EMIEngine::createLua() {
 	return new Lua_V2();
 }
 
+const char *EMIEngine::getUpdateFilename() {
+	if (getGamePlatform() == Common::kPlatformWindows && !(getGameFlags() & ADGF_DEMO)) {
+		switch (getGameLanguage()) {
+			case Common::FR_FRA:
+				return "MonkeyUpdate_FRA.exe";
+				break;
+			case Common::DE_DEU:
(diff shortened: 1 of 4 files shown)


andrews05:
Is it really a problem to support the patch file too? This would make it easy for versions that may include the patch already, or in my case, running from an existing installation which has already been patched. Not that it's hard to download the update, it would just be nicer if it wasn't necessary.

somaen:
It is a support problem, much easier to simply require it, as long as we don't know exactly which versions are what.

andrews05:
Would it be legal to distribute the patch.m4b file? Given that it's a free component of the update.

somaen:
No, it's free, but not freely distributable in all cases. (See the EULA, IIRC there's something about the package having to be distributed complete and unchanged). Which in turn means that it's fine for us to mirror the patch on our website, but problematic to bundle it with ResidualVM (GPL probably has issues too), and definitely a huge problem to distribute only parts of it.

For similar cases, see how Microsoft's corefonts, Adobe's Flash, or most binary blob drivers are installed by various distros: They are downloaded verbatim by some bootstrapping package, and then any necessities for handling them is performed locally.

I guess we could do something similar here, i.e. download the patch when it's missing, however I don't WANT us to do something like that, as it needlessly complicates a lot of stuff to suddenly handle anything related to networking/web/etc across a bunch of platforms.

PR #814, opened 2014-01-10
GRIM: Use exceptions instead of jumps

klusark:
This probably won't go in, but it was interesting figuring out that it just uses setjmp and longjmp like c++ exceptions.

This lets us get rid of some forbidden symbols, but ScummVM does not allow exceptions, so we probably can't use this. I mainly wanted to start some discussion on if there is any point in doing this.

Also, looking at lua 5.1 it uses exceptions in C++, but setjmp and longjmp for C.


aquadran:
@somaen rule followed by ScummVM setup, some platforms don't support exceptions. We are not limited.

aquadran:
@klusark when you will have some time, can you update it?

klusark:
@aquadran I've updated it.

PR #820, opened 2014-01-15
RFC: GRIM: Allow changing opengl modes on the fly

klusark:
If the platform supports both the new OpenGL with shaders and the old OpenGL we want to be able to switch between them at run time. This PR will let the user choose which OpenGL mode they want to use.

More changes will be needed to the configure script in order to actually build with both I think. I've only tested it on Windows by manually enabling shaders.

I added the "shaders" option. Maybe a better name is in order? We already have "use_arb_shaders" as well.


bgK:
Mostly? superseded by #1219

Issue #796, opened 2013-12-22
GRIM: New german Version with different files


nudgegoonies:
Hi,
i tried out the latest daily build and it print outs following messages:

```
WARNING: 'vox0003.lab' may be corrupted. MD5: '5bda9a36181c1fea368491b842c11efe'!
WARNING: 'vox0000.lab' may be corrupted. MD5: 'e884221099ba62dacf04a35be8428419'!
WARNING: 'movie03.lab' may be corrupted. MD5: '85a720936044f99ea7c08522606349b8'!
WARNING: 'movie01.lab' may be corrupted. MD5: 'b5a67c98be256eddaf731b39e0ceb8bd'!
WARNING: 'local.lab' may be corrupted. MD5: '27c961c16f9eaf0260781883150e5611'!
```

I use the german budget release on 2 CDs in a double case with documentation on CD. It has the ISBN 4012160461009. It is detected by residualvm as grim-win-de.
Regards,
Andreas

TobiX:
Shouldn't this be fixed after #845 was merged?

PR #845, opened 2014-03-23
GRIM: Fix MD5 sums for german versions.

TobiX:
These are the verified, correct checksums for most german disc versions
(after removal of the "copy protection" that is on some discs).

Please the the original MD5 forum thread for verification: http://forums.residualvm.org/viewtopic.php?t=215 - Here user desperadoduck, M[u]ddy and mark verify those MD5 sums as correct

Additionally, I can confirm those MD5 sums match with my own version. See also #796, where I try to explain how those incorrect MD5s ended up in the source code (probably a copy & paste error).


somaen:
Merging this, and crossing my fingers that it's correct.

PR #847, opened 2014-03-30
RFC: GRIM: Do not draw textures where all the texture coords are the same point. Fixes #58

### LINKED ISSUE #58: GRIM: Incorrect hole punch color with TinyGL
In original game, the hole punch is dark grey.
With opengl, it is lighter grey (maybe a difference in ambient light, or my gamma...).
With TinyGL, it is blue.

(yeah, this _is_ a super-ultra-minor glitch... :) )

klusark:
Probably not the best way to do this, but it illustrates the issue. I probably haven't though though all the edge cases though.

Anyone with better graphics knowledge have an opinion here?


aquadran:
can we closed this one?

vpelletier:
I wrote pull request #1259 after finding your change (but not this pull request for some reason). Basically, my change makes TinyGL "regress" to the level of OpenGL (ie, the hole punch is now grey everywhere), and leaves open the question of fixing the data issue.

PR #848, opened 2014-03-30
RFC GRIM: Move more of the logic regarding GL/TinyGL-selection to gfx_base.

somaen:
This is intended as a proposal to some changes in #820, it moves the checks for what driver to use if the users choice is not compiled in to gfxBase, and thus the relevant ifdefs.I f all else fails, we will fallback to TinyGL-rendering in the factory-function this way.

I have not tested this with the shaders-driver, and just a minimal test was done with the other two.

Another approach entirely would be to move even the ConfMan-check to GfxBase, and thus make grim.cpp completely oblivious about the selection.

All in all, this should leave us with a tad cleaner, and more flexible way to set up the drivers.

Thoughts, and testing with the GLES-backend is very welcome.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -157,7 +157,7 @@ void registerDefaults() {
 
 	// Graphics
 	ConfMan.registerDefault("fullscreen", false);
-	ConfMan.registerDefault("soft_renderer", false);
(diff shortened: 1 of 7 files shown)


bgK:
Superseded by #1219

PR #952, opened 2014-07-05
WIP: GRIM: Use Vertex Buffer Objects for meshes, when supported

klusark:
This greatly reduces the number of OpenGL calls needed each frame.

I'm not really happy with the state of the creation code right now, and additionally it is very similar to the code from the OpenGL Shaders backend, so it could potentially be merged somehow.

There are also still a few issues, for example on Eva

screenshot 2014-07-04 18 04 10

### DIFF
--- a/engines/grim/gfx_base.h
+++ b/engines/grim/gfx_base.h
@@ -65,6 +65,17 @@ public:
 	Texture *_texture;
 };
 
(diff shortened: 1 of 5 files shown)


aquadran:
we have OpenGL shader renderer now

PR #1050, opened 2014-08-18
RFC: EMI: Check for collisions when using updateWalk().

JoseJX:
This forces a check for collisions when updateWalk is called. This fixes Timmy running through the legs of Guybrush on the beach in the set mib. I think it fixes the logs in the flume puzzle as well, but I need to finish with that PR first to be sure since they're not visible right now.

### DIFF
--- a/engines/grim/actor.cpp
+++ b/engines/grim/actor.cpp
@@ -1426,6 +1426,21 @@ void Actor::updateWalk() {
 
 	dir = destPos - _pos;
 	dir.normalize();
+
+	// If we're not in a collision mode an walking, turn it on...
+	CollisionMode mode = _collisionMode;
+	if (_collisionMode == CollisionOff) {
+		mode = CollisionSphere;
+	}
+
+	// Find the amount we're walking
+	Math::Vector3d newWalkAmt = dir * walkAmt;
+	foreach (Actor *a, g_grim->getActiveActors()) {
+		// Don't move if we've collided
+		if(handleCollisionWith(a, mode, &newWalkAmt)) {
+			return;
+		}
+	}
 	_pos += dir * walkAmt;
 }
 


chkr-private:
This bug is still present:
- as described in this PR, when Timmy moves on the beach of Monkey Island, there is no collision with Guybrush
- another instance of the bug is #1153 

#1108 only fixed the issue if an actor is moved via Actor::setPos().

Unfortunately, the proposed bug fix has the drawback that the moving actor does not try to avoid the  object it collides with. This has the following two consequences:

1. if Timmy tries to run into the direction of Guybrush (but can't because he would collide), then he hops up and down until another walk attempt with a random direction is initiated by the lua code and allows him to move

2. when testing #1108 something similar happens

- first place Guybrush on the right side of the tourist in front of the statue and try to talk with the tourist
- Guybrush tries to move to the left side (there is a fixed point he is supposed to walk to), but he collides with the tourist
- regardless if the code would avoid walking at all (as done in this patch) or with a modified patch where the actor would walk at least the returned amount from handleCollisionWith(), Guybrush would not move, but keep playing the walking sequence forever

engines/grim/actor.cpp
```diff
@@ -1426,6 +1426,21 @@ void Actor::updateWalk() {
...
+	CollisionMode mode = _collisionMode;
+	if (_collisionMode == CollisionOff) {
+		mode = CollisionSphere;
+	}
+
+	// Find the amount we're walking
+	Math::Vector3d newWalkAmt = dir * walkAmt;
+	foreach (Actor *a, g_grim->getActiveActors()) {
+		// Don't move if we've collided
+		if(handleCollisionWith(a, mode, &newWalkAmt)) {
```
somaen: Missing a space after if.

PR #1109, opened 2014-10-15
RFC/Discussion: GRAPHICS: Proof-of-concept: Render (well, blit) TinyGL using OpenGL 

somaen:
This was something I threw together quite rapidly as a proof-of-concept, and is thus not meant for merging, but rather for discussion.

The point here, is to allow for rendering the result of TinyGL using OpenGL (which should be doable with almost any OpenGL(ES)-renderer in theory.

A few points to note:
- I did not do anything other than copy around the Overlay-code (which is now horribly duplicated). I think we can make a struct of the duplicated variables, and thus reuse the existing functions for this, instead of duplicating, passing a flag to the functions to decide whether to work on _overlayX, or _gameScreenX.
- I did not do anything with _gameScreenDirty, as this is a proof-of concept.
- I did not expose a way to set the filter for rendering. (GL_LINEAR vs GL_NEAREST)
- This does not actually handle any aspect/scaling issues, as that depends on adding the GUI-flags etc from #1080.
- It should in theory be a workable solution for drawing TinyGL with arbitrary scaling on any machine that has GLES 1.1 or OpenGL 1.2 available (given that the aspect-flag and some additional code to keep within it is added).
- No code to detect the availability of OpenGL was added, but that should be doable.
- No fallback solution for the cases where absolutely NO OpenGL is available is provided here yet, (indeed this code doesn't even allow for that to happen, but this is, once again, just a proof-of-concept).

Comments on the general idea or overall approach is very welcome.

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -50,15 +50,9 @@ SurfaceSdlGraphicsManager::SurfaceSdlGraphicsManager(SdlEventSource *sdlEventSou
 	:
 	SdlGraphicsManager(sdlEventSource),
 	_screen(0),
-	_overlayVisible(false),
-	_overlayscreen(0),
-	_overlayWidth(0), _overlayHeight(0),
-	_overlayDirty(true),
 	_screenChangeCount(0)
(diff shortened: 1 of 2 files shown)


aquadran:
it's already part of scummvm now

Issue #992, opened 2014-08-03
EMI: No path found for guybrush when getting the boulders on monkey island


klusark:
To reproduce:
Jump to set pac
Target the boulders and press p
It should walk guybrush to the rocks. Instead it does nothing.

Caused by this commit 3ec234757f0cf365d86e40f485a68fcc5f6115dd

After reverting that commit it works fine.

chkr-private:
Fixed by commit 39b0172f721c4307ed3bdd36ce0c54b654c1e0fc.

PR #1127, opened 2014-12-12
EMI: Relax height check for bridges between sectors

chkr-private:
- fixes a path finding issue (#992)


chkr-private:
@Akz-: I have checked the problematic sectors: they are indeed not parallel. Even worse, one sector isn't even a plane. Implementing 3D intersection for non-plane polygons would be complicated. Given the minor benefit of such an implementation, I believe that the best way to proceed would still be just to relax the height check.

somaen:
I have a hunch we keep coming back to this, so I'd appreciate a comment with a reference to issues here before merging.

chkr-private:
@somaen: I have added some comments to the commit. OK to merge?

PR #1143, opened 2014-12-31
GRIM/EMI: Limit frequency of luaUpdate()

chkr-private:
- if luaUpdate() is called more than once during the same millisecond
  then delay the next call by one millisecond
- this avoids the issue that _frameTime is set to 0 which would
  cause getPerSecond() to return 0 for any given rate
- if getPerSecond() returns 0 then that would cause dist to be 0 in walkForward() which would cause the walk animation to stop (even if the actor is supposed to keep walking)
- if that happens, e.g. Guybrush would walk quite erratically

### DIFF
--- a/engines/grim/grim.cpp
+++ b/engines/grim/grim.cpp
@@ -798,15 +798,22 @@ void GrimEngine::mainLoop() {
 
 				handleControls(type, event.kbd);
 
-				// Allow lua to react to the event.
-				// Without this lua_update switching the entries in the menu is slow because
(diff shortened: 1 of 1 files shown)


klusark:
The only case that this would happen is on user input. The user input part needs to be there for the grim menus as they are run as a single frame per key press instead of 60 frames per second. Maybe we can just disable that call for EMI? Would EMI need that call anywhere?

chkr-private:
I updated the pull request with an implementation of klusark's proposal.

PR #1157, opened 2015-01-12
GRIM: Make patch file reads alignment safe

joostp:
This changes the unsafe pointer reads/writes to use READ_UINT32() / WRITE_UINT32().
Since there are no 64-bit versions of these functions, the 64-bit codepath was eliminated.

I don't know how big the performance gains here are for splitting this into 4-byte chunks, the code would definitely be more readable if it were changed into a simple 1-byte xor...

### DIFF
--- a/engines/grim/patchr.cpp
+++ b/engines/grim/patchr.cpp
@@ -208,26 +208,18 @@ uint32 PatchedFile::read(void *dataPtr, uint32 dataSize) {
 			_diffCopy -= readSize;
 
 			//Read data from diff as blocks of size _kDiffBufferSize,
-			// then xor original data with them in groups of 4 or 8
-			// bytes, depending on the architecture
+			// then xor original data with them in groups of 4 bytes
 			while (readSize > 0) {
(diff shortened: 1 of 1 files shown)


aquadran:
@YakBizzarro: can you comment that part of 32/64 bit, why is that?

YakBizzarro:
My idea was to use the larger register size to reduce the number of the iterations. Maybe it's only a premature optimization, since I never looked at the resulting assembly or benchmarked it. Anyway, keep in mind that a very large part of the file that need to be patched, must pass through that cycle. Untouched part of a file are just xored with zero, which are greatly compressed by gzip, so the number of cycles is quite high

aquadran:
I'm not sure if it's worth. At this point aligned is more important, I'm thinking about to merge it. We can bring this back (after merge 64 macros) if it's indeed faster, if not it make no sense at all to complicate code.

YakBizzarro:
Ok. I'll reintroduce the 64bit codepath if I can prove that it's a real optimization ;)

Issue #1158, opened 2015-01-12
EMI: Crash due to unknown pixelformat


joostp:
Trying to run EMI on latest master on an SGI Fuel running IRIX using the OpenGL renderer.
The "LucasArts" intro starts okay, but then when I press ESC to skip to the game (so it should load the intro movie that starts with guybrush's face) ResidualVM errors out with:
"Unknown pixelformat: Bpp: 2 RBits: 5 GBits: 5 BBits: 5 ABits: 0 RShift: 0 GShift: 5 BShift: 10 AShift: 0!"

OpenGL config for this machine:

INFO: OpenGL Vendor: SGI
INFO: OpenGL Renderer: VPRO/B/32
INFO: OpenGL Version: 1.2 Irix 6.5
INFO: OpenGL Red bits: 8
INFO: OpenGL Green bits: 8
INFO: OpenGL Blue bits: 8
INFO: OpenGL Alpha bits: 8
INFO: OpenGL Z buffer depth bits: 24
INFO: OpenGL Double Buffer: 1
INFO: OpenGL Stencil buffer bits: 8

Same error doesn't occur when using tinygl (but it has its own set of problems).

joostp:
#1159 fixes this.

joostp:
This is fixed by 7bdd1af0d73e5c957248ff41ae458807aa90cad9

PR #1159, opened 2015-01-13
EMI: Add support for 555 pixelformat to GfxOpenGL::prepareMovieFrame() (...

joostp:
...for SGI VPRO)

Fixes issue #1158 (thanks @somaen )


bgK:
Shouldn't VideoDecoder::setDefaultHighColorFormat be called to always use the same pixel format ?

PR #1167, opened 2015-02-02
GRIM: Fix compile warnings in codec48.cpp

chkr-private:
This PR fixes a bunch of compile warnings of the following type:

<pre>warning: cast from type ‘const byte* {aka const unsigned char*}’ to type ‘uint32* {aka unsigned int*}’ casts away qualifiers [-Wcast-qual]</pre>


Botje:
:+1:

PR #1170, opened 2015-02-08
EMI: Allow saving in-game on the PS2 version

d356:
Change the memory card id and the reported free space so that the game is able to save.

There are some remaining issues:

* When finished loading a game, the game displays “Save Successful.” This seems to be just a visual bug.

* The game saves without *.gsv file extensions, resulting in filenames such as “efmi000”.  This prevents loading one of these saves from ResidualVM’s menus.

* Save thumbnails are displayed incorrectly.

### DIFF
--- a/engines/grim/emi/lua_v2.cpp
+++ b/engines/grim/emi/lua_v2.cpp
@@ -477,14 +477,19 @@ void Lua_V2::ThumbnailFromFile() {
 		return;
 	}
 	int index = (int)lua_getnumber(texIdObj);
-	const char *filename = lua_getstring(filenameObj);
+	Common::String filename(lua_getstring(filenameObj));
+	
+	if (g_grim->getGameType() == GType_MONKEY4 &&
+		g_grim->getGamePlatform() == Common::kPlatformPS2) {
+		filename += ".gsv";
+	}
 
 	int width = 256, height = 128;
 
 	SaveGame *savedState = SaveGame::openForLoading(filename);
 	if (!savedState || !savedState->isCompatible()) {
 		delete savedState;
-		warning("Lua_V2::ThumbnailFromFile: savegame %s not compatible", filename);
+		warning("Lua_V2::ThumbnailFromFile: savegame %s not compatible", filename.c_str());
 		lua_pushnil();
 		return;
 	}
@@ -503,7 +508,7 @@ void Lua_V2::ThumbnailFromFile() {
 	Bitmap *screenshot = new Bitmap(buf, width, height, "screenshot");
 	if (!screenshot) {
 		lua_pushnil();
-		warning("Lua_V2::ThumbnailFromFile: Could not restore screenshot from file %s", filename);
+		warning("Lua_V2::ThumbnailFromFile: Could not restore screenshot from file %s", filename.c_str());
 		delete screenshot;
 		delete[] data;
 		delete savedState;
--- a/engines/grim/emi/lua_v2.cpp
+++ b/engines/grim/emi/lua_v2.cpp
@@ -481,7 +481,7 @@ void Lua_V2::ThumbnailFromFile() {
 	
 	if (g_grim->getGameType() == GType_MONKEY4 &&
 		g_grim->getGamePlatform() == Common::kPlatformPS2) {
-		filename += ".gsv";
+		filename += ".ps2";
 	}
 
 	int width = 256, height = 128;
--- a/engines/grim/emi/lua_v2.cpp
+++ b/engines/grim/emi/lua_v2.cpp
@@ -522,8 +522,10 @@ void Lua_V2::ThumbnailFromFile() {
 
 void Lua_V2::GetMemoryCardId() {
 	// 0 - No mem card
-	lua_pushnumber(0);
-	warning("GetMemoryCardId: Currently just returning 0");
+	// 1 - Not formatted
+	// 2 - Not enough space
+	// 3 - Error occured
+	lua_pushnumber(4);
 }
 
 void Lua_V2::SetReplayMode() {
--- a/engines/grim/lua_v1.cpp
+++ b/engines/grim/lua_v1.cpp
@@ -39,6 +39,7 @@
 #include "engines/grim/bitmap.h"
 #include "engines/grim/font.h"
 #include "engines/grim/gfx_base.h"
+#include "engines/grim/localize.h"
 
 #include "engines/grim/lua/lauxlib.h"
 #include "engines/grim/lua/luadebug.h"
@@ -554,6 +555,7 @@ void Lua_V1::SubmitSaveGameData() {
 		error("Cannot obtain saved game");
 	savedState->beginSection('SUBS');
 	int count = 0;
+	Common::String localized;
 	for (;;) {
 		lua_pushobject(table);
 		lua_pushnumber(count);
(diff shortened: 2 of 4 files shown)


somaen:
The file-suffix can probably be handled in Save/Load by adding code wrapped in if (g_grim->getGamePlatform() == Common::PlayStation2) to add the .gsv

Not super sure if that's the exact check, but should be trivial to look up. Does the load menu work fine too?

engines/grim/lua_v1.cpp
```diff
@@ -497,8 +497,14 @@ void Lua_V1::GetSpeechMode() {
 }
 
 void Lua_V1::GetDiskFreeSpace() {
+	if(g_grim->getGameType() == GType_MONKEY4 &&
+		g_grim->getGamePlatform() == Common::kPlatformPS2) {
+		//wants more than 600 KB
+		lua_pushnumber(700);
+	} else {
 	// amount of free space in MB, used for creating saves
 	lua_pushnumber(50);
```
somaen: This should be indented when adding the else-clause around it.

engines/grim/lua_v1.cpp
```diff
@@ -497,8 +497,14 @@ void Lua_V1::GetSpeechMode() {
 }
 
 void Lua_V1::GetDiskFreeSpace() {
+	if(g_grim->getGameType() == GType_MONKEY4 &&
```
somaen: We prefer a space between the if and the ( for consistency (and having it consistent helps with searching in the code base for instance).

somaen:
Thanks a lot for this, I've looked at it, and aside from the two minor code formatting comments above, it looks good to merge.

d356:
Thanks, I've corrected the white-space and I will look into fixing the file extension issue.

engines/grim/emi/lua_v2.cpp
```diff
@@ -522,8 +527,10 @@ void Lua_V2::ThumbnailFromFile() {
 
 void Lua_V2::GetMemoryCardId() {
 	// 0 - No mem card
-	lua_pushnumber(0);
-	warning("GetMemoryCardId: Currently just returning 0");
+	// 1 - Not formatted
+	// 2 - Not enough space
+	// 3 - Error occured
```
somaen: Out of curiosity, is any other number a valid memory card id?

engines/grim/lua_v1.cpp
```diff
@@ -497,8 +497,14 @@ void Lua_V1::GetSpeechMode() {
...
 
 void Lua_V1::GetDiskFreeSpace() {
+	if (g_grim->getGameType() == GType_MONKEY4 &&
+		g_grim->getGamePlatform() == Common::kPlatformPS2) {
+		//wants more than 600 KB
+		lua_pushnumber(700);
+	} else {
 	// amount of free space in MB, used for creating saves
-	lua_pushnumber(50);
+		lua_pushnumber(50);
```
somaen: This makes me think that an even simpler fix is possible, i.e. just returning 700 for both games. Sure Grim Fandango will then think that you have 700 MB free (who hasn't these days?), but overall, that should fit both.

Then again, this solution makes the behaviour in the two games clear, so it doesn't hurt.

engines/grim/emi/lua_v2.cpp
d356: No other numbers seem to be specifically checked for.  I'm not sure what a valid id would look like on a real console but returning "4" is apparently enough to let saving work.

engines/grim/lua_v1.cpp
d356: Ok, I can revise the patch so that it returns 700 for both games.

d356:
By description string I mean what is shown in the Load screen in ResidualVM's menus. For example instead of something a human would write as the name for their save like "0. Wedding Ship" it displays "0. /txtsavwed/".

somaen:
That list is filled by listSaves in detection.cpp IIRC. (And the strings are originally saved by SubmitSaveGame in Lua_V1.)

However, that string does look quite like a localization string for EMI. I wonder, would it be very wrong (tm) to force the first string to be saved localized? (i.e. calling g_localizer->localizeString for the very first string only, and only for EMI PS2) I can't see any other option though, as the localization table isn't really available from the menu? Perhaps extending the save game format to actually store the description in ResidualVM friendly localized form in a section before the actual save (which would change the save game format, and warrant a bit of modifying of the save-game loading code).

somaen:
As a sidenote, the current solution has a little quirk:
Savegames are currently named efmi000.gsv no matter what the game target is set to, which means that EMI PS2 picks up on PC/Mac saves too (which of course are NOT compatible, as the lua state is saved in the save games, and the scripts differ, PC/Mac savegames might be compatible with eachother, but not with PS2).

So, there are two potential things that should (perhaps both) be fixed:
1: PS2 Savegames should contain a tag telling that they are in fact PS2-savegames.
2a: Savegames should get filenames that follow the targetname from the launcher. (Which also means modifying listSaves in detection.cpp accordingly. 
OR
2b: Savegames should at the very least get a different name-scheme when coming from PS2-versions, to avoid the launcher and game from picking up on Mac/PC-saves when running the PS2-version.

somaen:
Another little problem: If a savegame is invalid, this does not allow for showing so, I tried modifying the Thumbnail read code in engines/grim/emi/lua_v2.cpp, which made the save games incompatible. The result was that the slots were listed as empty, and no warning was given about potentially overwriting the save game. (I did get the empty slot warning though).

somaen:
Found the fix for the thumbnails: The Mac/PC versions use a 256/128 texture, with only 160/120 used inside it, where the PS2 version uses the full texture.

If you change the width and height in EMIEngine::storeSaveGameImage (engines/grim/emi/emi.cpp) to 256/128, and the for-loop below accordingly, you'll get the correct thumbs stored. (Of course this needs an if on whether we are actually in a PS2-savegame).

Botje:
I think naming the PS2 save files as `efmiXXXX.ps2` is acceptable, but I don't know how we can plumb the registered games into the meta-engine.

d356:
To handle the localization string issue I decided to use beginSection() to add a new section to the save file.  This seemed like the least invasive thing to do.

I decided to change the extension of the save files to *.ps2.  I think this is the simplest way to handle the incompatibility issues.

I haven’t changed anything regarding somaen’s “Another little problem:” comment. Maybe it’s no longer an issue?

Since somaen already has the fix for the thumbnail issue I figured he can just commit his fix after this pull request is merged.

I’m not sure what “plumb the registered games into the meta-engine.” means.  Could you elaborate?

klusark:
@d356 The meta engine is the class which contains meta data about the engine and creates it based on detection rules. It also has other features such as listing save games in the main interface so you can directly load a save instead of having to load into the game first. Here's the code that does it
https://github.com/residualvm/residualvm/blob/master/engines/grim/detection.cpp#L582

somaen:
@d356: Feel free to commit the thumbnail fix too if you want.

somaen:
What happens if the save is from the PC-version now?

d356:
@klusark Thanks for the explanation.

@somaen Saves from the PC version don't show in the launcher for the PS2 version or the other way around, so it should be impossible to accidentally load an incompatible save. You're right that it shouldn't be an exact equality, I have fixed that now.

Issue #754, opened 2013-09-30
GRIM: Strange lighting on Meche's face when first talking to Manny


RyanJWilliams:
![residualvm 2013-09-29 22-19-38-77](https://f.cloud.github.com/assets/5575257/1236186/6edf08a0-29b9-11e3-9b44-d259d667db6b.png)

DxDiag: http://pastebin.com/4QKSxbvK

PR #1177, opened 2015-03-15
GRIM: Assorted lighting fixes

vpelletier:


vpelletier:
In reply to in-line comments from @somaen , as those messages were made hard to access because of a push-force on my repository - my bad.

TinyGL,GRIM: Store spot direction in eye coordinates.

- Does this follow the expected OpenGL behaviour?

  Yes: https://www.opengl.org/sdk/docs/man2/xhtml/glLight.xml GL_SPOT_DIRECTION:

```
The spot direction is transformed by the upper 3x3 of the modelview matrix when glLight is called, and it is stored in eye coordinates.
```

- Also, you should probably split this into two commits, one TINYGL-commit, and one GRIM-commit.

  Ok, will do.

TinyGL: Use norm_position for directional lights.

- Does this follow the OpenGL spec?

  I do not know if OpenGL specifies this. ```d``` is used in dot-products to compute angles, which AFAIK requires normalized vectors to give any meaningful result. For directional lighting, vector can be normalized once only when setting light "position" (direction, actually), as there is no distance computation to do (no attenuation), as opposed to positional lights. Also, norm_position was not used anywhere. I suspect this slight bug was not noticed because likely all lights are provided already normalized.

TinyGL: Always normalize object-to-light vector.

- Is there a reason why 1E-3 was originally used?

  I have no idea, but as per previous case, this value is used for dot-products, where non-normalized would cause bogus results. The way I understand this code, it's basically duplicating the normalize() method with the advantage of computing vector's norm once (before normalization, of course).

GRIM: Divide color by 15 once only.

- This / 1.3 seems to be gone below. Not really sure why OpenGL and TinyGL differ here though.

  I believe it is a remain from lighting code evolution. Without it and after my changes, lighting is extremely close to OpenGL result, to my eyes. Only some aliasing on edges and textures and some dithering differs.

GRIM: Resynchronise TinyGL setupLight with OpenGL's.

- Preferably this should be contained in the commits that originally diverged, if they are part of your commit-chain.

  They are not part of this chain, sadly, but much older commits from my original work on OpenGL lighting. I can look this/these commit(s) up and split this one with consistent commit messages, if you prefer.

GRIM: Get rid of specular lighting.

- Does this affect EMI in any way?

  Good point, it likely does. I'll give it a try, but I'm much less comfortable with it than with GRIM.

vpelletier:
For EMI, at least the two first interactive scenes (ship deck and governor house) look very similar to original (opengl renderer on XP in virtualbox), with no obvious specular lighting.

vpelletier:
Updated & simplified patchset.

A lot of changes:
- `GRIM: Get rid of specular lighting.` now reverts my former variable renaming, which in turn simplifies a lot `GRIM: Resynchronise TinyGL setupLight with OpenGL's.` to the point that I merged it into the GRIM part of `TinyGL,GRIM: Store spot direction in eye coordinates.`, as it only fixes spotlight case.
- Drop `TinyGL: Use norm_position for directional lights. ` as it should be a no-op anyway.
- Drop `TinyGL: Always normalize object-to-light vector.` as floating point division is slower than multiplication (leaving one division and 3 multiplications instead of 3 divisions), and 1E-3 is likely to be related to FLT_EPSILON (1E-5 ?), as float equality tests are bad.
- `TinyGL,GRIM: Store spot direction in eye coordinates.`: Split and TINYGL part simplified.
- `GRIM: Divide color by 15 once only.` changelog reversed (main feature of this commit is to get rid of TinyGL rederer-only divisor, reducing the number of floating point operations is a bonus).
- Split reflectance and attenuation patch into two.
- Re-checked spotlight attenuation, and actually there is ! So I removed special casing for spotlights.

Also, I generally made commit messages (much) more verbose.

vpelletier:
> - Re-checked spotlight attenuation, and actually there is ! So I removed special casing for spotlights.

...or maybe not. Leaving quadratic attenuation enabled "breaks" (as in, a bit too dark) at least sets `al` and `do`. I'm lost.

vpelletier:
Even weirder: looking into #754 (lighting issue when talking with Meche in set "mo"), there seem to be no attenuation here too, although this set only contains "omni" and "direct" light types. But even when removing attenuation, the shadow over Meche's mouth doesn't go away (it merely becomes lighter). I'm still puzzled as to how the game decides to apply attenuation or not.

vpelletier:
Would this pull request be accepted if I drop the 2 last commits (the ones about attenuation) ?

vpelletier:
Rebased on current master.
Added one more attenuation fix, for shader renderer (I forgot OpenGL default constant attenuation).

vpelletier:
Rebased again. Any blocker in this patch set ?

Botje:
Other than that, looks good to me. Although I don't have much experience with grim

vpelletier:
Amended "TinyGL: Store spot direction in eye coordinates" so point being transformed is more visible. Also, rebased over current master.

Botje:
Okay. Any final objections to merging this?It looks like the worst case is that some grim models are overly lighted.

aquadran:
fine by me

PR #1179, opened 2015-03-20
GRIM: Side texture support for the widescreen mod

Botje:
See http://forums.residualvm.org/viewtopic.php?f=2&t=687&start=15#p3765.
Is there a chance of getting this merged, modulo some cleanup and/or refactoring?


klusark:
:+1: :+1: on the concept

aquadran:
ok

Botje:
I addressed all review comments, rebased on top of master, and added support for TinyGL.
Comments welcome!

bgK:
I'm a bit worried about the added dependency between this feature, the GRIM engine and the SDL backend. Shouldn't this have been implemented completely in the GRIM engine? I realize it would have required a lot more work.
Maybe a comment should be added stating that this is a hack and should be reimplemented in the engine once kSupportsArbitraryResolutions is done.

Botje:
@bgK you're right. I chose this route because there was pushback against decoupling the render resolution from 640x480 in grim.

aquadran:
I agree with @bgK we can't access SDL backend from grim engine directly, this has to be done by System API, if you need some new function for this we may add new one.

Botje:
The branch now adds and uses a `suggestSideTextures` method to System.
Is this acceptable?

engines/grim/grim.h
```diff
@@ -188,6 +188,10 @@ class GrimEngine : public Engine {
 	virtual void openMainMenuDialog() override;
 	void debugLua(const Common::String &str);
 
+#ifdef SDL_BACKEND
```
bgK: That ifdef should not be needed anymore

common/system.h
```diff
@@ -1295,6 +1298,21 @@ class OSystem : Common::NonCopyable {
...
+	 * 
+	 * The system must take ownership or the pointers will be freed
+	 * automatically.
+	 *
+	 * !!! ResidualVM specific method: !!!
+	 *
+	 * @param left			Texture to be used on the left
+	 * @param height		Texture to be used on the right
+	 */
+	virtual void suggestSideTextures(Common::ScopedPtr<Graphics::Surface>& left,
```
bgK: I'm a bit unsure about passing Common::ScopedPtr references as parameters. It works fine, but feels a bit strange to me.

bgK:
Fine to merge with me, once the ifdef and build are fixed. :+1:

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -569,7 +573,23 @@ void SurfaceSdlGraphicsManager::drawOverlayOpenGL() {
 	glPopAttrib();
 }
 
-void SurfaceSdlGraphicsManager::drawFramebufferOpenGL() {
+void SurfaceSdlGraphicsManager::drawSideTexturesOpenGL() {
+	if (_fullscreen && _lockAspectRatio) {
+		const Math::Vector2d nudge(1.0/float(_screen->w), 0);
```
aquadran: formating

backends/graphics/surfacesdl/surfacesdl-graphics.h
```diff
@@ -112,6 +113,7 @@ class SurfaceSdlGraphicsManager : public SdlGraphicsManager, public Common::Even
 	virtual int16 getOverlayHeight() { return _overlayHeight; }
 	virtual int16 getOverlayWidth() { return _overlayWidth; }
 	void closeOverlay(); // ResidualVM specific method
+	virtual void setSideTextures(Graphics::Surface *left, Graphics::Surface *right);
```
aquadran: add note like line above

common/system.h
aquadran: @bgK can you explain?

backends/platform/sdl/sdl.cpp
```diff
@@ -490,3 +490,15 @@ Common::TimerManager *OSystem_SDL::getTimerManager() {
 	return _timerManager;
 #endif
 }
+
+bool OSystem_SDL::hasFeature(Feature f) {
```
aquadran: Can you implement this function and below in SDL graphics backend? I would avoid changes here.

engines/grim/grim.cpp
```diff
@@ -315,6 +319,7 @@ Common::Error GrimEngine::run() {
 	g_driver->setupScreen(640, 480, fullscreen);
 	g_driver->loadEmergFont();
 
+
```
aquadran: no need new line

engines/grim/grim.cpp
```diff
@@ -69,6 +71,7 @@
 
 #include "engines/grim/lua/lua.h"
 
+
```
aquadran: no need new line

backends/platform/sdl/sdl.cpp
```diff
@@ -490,3 +490,15 @@ Common::TimerManager *OSystem_SDL::getTimerManager() {
...
+	if (f == kFeatureSideTextures)
+		return true;
+	return ModularBackend::hasFeature(f);
+}
+
+void OSystem_SDL::suggestSideTextures(
```
aquadran: I think we don't nee implement this for generic SDL

common/system.h
bgK: After a more thorough look at the memory management here, I think it is incorrect because Graphics::Surface pointers need to have their 'free' method called to release their pixel data.
I suggest either:
* Common::SharedPtr along with Graphics::SharedPtrSurfaceDeleter is used instead of Common::ScopedPtr.
* Or preferably a plain Graphics::Surface pointer is passed as a parameter to suggestSideTextures, and is freed in GrimEngine::setSideTextures, explicitly stating in a Doxygen comment that the backend is expected to copy the surface data.

Botje:
@aquadran: are you talking about just the `hasFeature` method or also the `suggestSideTextures` method?

Botje:
Per our IRC discussion, @aquadran is okay with the current solution.
The commit pushed last simply moves the method closer to a block of ResidualVM-specific code.

Any last comments on the current set of commits before I rebase against master?

engines/grim/grim.cpp
```diff
@@ -1374,6 +1378,30 @@ void GrimEngine::pauseEngineIntern(bool pause) {
...
+}
+
+void GrimEngine::setSideTextures(const Common::String &setup) {
+	if (! g_system->hasFeature(OSystem::kFeatureSideTextures))
+			return;
+	Graphics::Surface *t1 = loadPNG(Common::String::format("%s_left.png", setup.c_str()));
+	Graphics::Surface *t2 = loadPNG(Common::String::format("%s_right.png", setup.c_str()));
+	g_system->suggestSideTextures(t1, t2);
+	t1->free();
+	t2->free();
```
bgK: Please also delete the surfaces.

aquadran:
:+1:  I'm fine with merge.

engines/grim/grim.cpp
Botje: Gah. too many small changes. Good catch.

Botje:
This code is rebased against master.
With the addition of side textures I did notice a black line around the game screen, I'm still trying to figure out why this is.

bgK:
Is that due to the textures not being fully used, in combination with bilinear filtering and GL_TEXTURE_WRAP having no effect on the padded sides?

Botje:
Right you are. I changed `Graphics::Texture` to make use of NPoT textures where possible.
Now it looks great!
screen shot 2015-08-18 at 14 22 37

Botje:
If no-one objects, I'll merge this on Friday.

klusark:
LGTM :+1:

PR #1180, opened 2015-03-21
RFC Enable GRIM glsl lighting

vpelletier:
This patch set enables lighting code in actor vertex shader.
This is my first stab at shaders, so please carefully review and tell me what I'm doing wrong.
I tried to change as little CPU as possible, because I'm afraid of all the special-casing for GRIM vs. EMI.
I did move attenuation computation to shader, though, which required me to update EMI shader. Was this a good idea ? My idea is that shader needs to multiply color with per-vertex intensity anyway, so rather than starting at intensity 1.0 and reducing it with normal-to-light, distance and cone attenuation, I should be able to just start at intensity and do the same for the same cost. I also tried to code the shader defensively whenever a divide-by-zero is possible, although I would tend to trust game data.


vpelletier:
Now with last commit split. The same shader works for GRIM with and without the move-CPU-mutliplications-to-GPU commit.

Botje:
Looks good!
I had to change some integer constants to floats (ie `0` -> `0.0`) in `emi_actor.vertex` and `grim_actor.vertex` in order to get ResidualVM to run on Android, though.

vpelletier:
Added (hopefully) all missing `.0`s, and added a trivial shadow fix for EMI.

PR #1182, opened 2015-03-24
GRIM: SHADERS: Enable GL_BLEND when texture has transparency.

vpelletier:
Fixes, likely among others, the tie rope in set "al".

### DIFF
--- a/engines/grim/gfx_opengl_shaders.cpp
+++ b/engines/grim/gfx_opengl_shaders.cpp
@@ -1203,7 +1203,7 @@ void GfxOpenGLS::selectTexture(const Texture *texture) {
 	GLuint *textures = (GLuint *)texture->_texture;
 	glBindTexture(GL_TEXTURE_2D, textures[0]);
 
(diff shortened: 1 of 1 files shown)


bgK:
The OpenGL renderer seems to do alpha testing. Not sure if doing blending instead has much of an impact.

vpelletier:
It does fix the tie rope for me, at least. No idea if it is the right way, though.

vpelletier:
I just noticed I can see through Glotis' mouth in set "ga". I didn't check if it is due to this commit.

bgK:
The alpha test approach seems to get both the rope and Glotis right.
https://gist.github.com/bgK/0308b600a10359cb1489

vpelletier:
I confirm your patch works better. Should I close this pull request and let you push your patch ?

bgK:
Done in https://github.com/residualvm/residualvm/commit/d1b72c7975b6697b47bb7bc35a73d3b20d89af26

PR #1183, opened 2015-03-25
Primitives

vpelletier:
Fix primitives rendering with TinyGL and Shaders.


vpelletier:
True, and so far I'm clueless as to how it should happen or what should be fixed.
FWIW, slapping more presentBuffer fixes primitives more (truncated "up" arrow and fresque dimming in save/load menus, at least, and likely the iris as well), but at the cost of a x2 cpu usage increase when using Copal's computer.
Should I drop the tinygl part of this pull request ? (ie, is the shader part worth anything ?)

aquadran:
yes, you should drop tinygl part :)

Botje:
:+1:

Issue #1169, opened 2015-02-03
EMI: Pegnose has zero animation


monkeymadness:
The way to reproduce this is to throw the duck in the window without putting grease on the mat. If done so, Pegnose will glide across the floor with his arms and feet at his side.

![pegnose1](https://cloud.githubusercontent.com/assets/10478990/6026352/75a1380a-ab97-11e4-9007-c67b00f79b9d.png)
![pegnose2](https://cloud.githubusercontent.com/assets/10478990/6026360/7c226c94-ab97-11e4-9396-7cfa0b3e3386.png)
![pegnose3](https://cloud.githubusercontent.com/assets/10478990/6026366/8222f1fe-ab97-11e4-9404-f778b89832d0.png)
![pegnose4](https://cloud.githubusercontent.com/assets/10478990/6026371/8803c3c8-ab97-11e4-846f-fb8f3fc55627.png)

Version of the game: English version from a jewelcase
OS: Windows 7 64bit
Graphics Card: Intel HD Graphics Driver
Renderer: OpenGL
Version of ResidualVM: 0.3.0gitc5ccf4d

Save File:
![efmi008](https://cloud.githubusercontent.com/assets/10478990/6026390/afb31b08-ab97-11e4-8c74-06e8faef5ae3.png)

chkr-private:
Fixed by commit 409a6f7b95a3f86accbb0ba71e65ecd53a82c58b.

PR #1189, opened 2015-05-03
EMI: Start animation when "Loop" key is set

chkr-private:
This fixes the problem if the costume uses only key 3 ("Loop") but not
key 1 ("Play") for an animation. This happens e.g. for Pegnose's "run" chore (issue #1169).


klusark:
Are there any other cases like Pegnose, but work differently?

chkr-private:
@klusark: What I have checked so far:
- the problem, that "Loop" is used but not "Play", is limited to Pegnose's running animation
- the only other potentially critical combination would be that "Play" would have a later time than "Loop"  - but I guess the semantic would still be to start looping anyway...

Issue #1155, opened 2015-01-10
EMI: Some of the credits do not do not have the right color tint


monkeymadness:
The credits have usually around half stay a dark grey instead of the right color.

![weirdendcredits](https://cloud.githubusercontent.com/assets/10478990/5692970/29b7faa6-98c7-11e4-9cc5-5d076b2857d4.jpg)

Version of the game: English version from a jewelcase
OS: Windows 7 64bit
Graphics Card: Intel HD Graphics Driver
Renderer: OpenGL
Version of ResidualVM: 0.3.0gitae24374

chkr-private:
Fixed by commit be771ab4546ac589b81d7291c3156e79cbaf28a2.

PR #1190, opened 2015-05-03
GRIM/EMI: Re-use TextObjects with the same textID and layer

chkr-private:
Re-use existing TextObjects in MakeTextObjects() in order
to avoid a massive duplications in case the LUA code
calls MakeTextObject multiple times for the same text
but with a different color.

Note:
In EMI, the menus are using 2 TextObjects with the same textID
with different colors but with a different layer.

This fixes the issue that in the credits of EMI the colors
were off (issue #1155)  since there were multiple TextObjects for the same
textID but each with a different color (one per fade step).

### DIFF
--- a/engines/grim/lua.h
+++ b/engines/grim/lua.h
@@ -165,6 +165,7 @@ private:
 	int _translationMode;
 	unsigned int _frameTimeCollection;
 
+public:
 	int refSystemTable;
 	int refTypeOverride;
 	int refOldConcatFallback;
--- a/engines/grim/lua_v1_text.cpp
+++ b/engines/grim/lua_v1_text.cpp
@@ -107,12 +107,35 @@ void Lua_V1::MakeTextObject() {
 	if (!lua_isstring(textObj)) {
 		return;
 	}
-
-	TextObject *textObject = new TextObject();
-	const char *line = lua_getstring(textObj);
-
-	textObject->setDefaults(&g_grim->_blastTextDefaults);
 	lua_Object tableObj = lua_getparam(2);
+
+	int layer = 0;
+	if (lua_istable(tableObj)) {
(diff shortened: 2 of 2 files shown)


klusark:
I'm not getting how this works. We have two text objects that only share their text content and layer, but we are representing them by the same text object, correct? But, if the colour differs between them, we throw out the second one? That seems like odd behavior to me.

chkr-private:
@klusark:
My first idea was: for one textID (one actual string) just use one text object. If a text object already exists when MakeTextObject is called, then return this text object to the LUA code instead of creating a new one.

This would have avoided the original issue that for each color fade step in the credits there would have been a new text object created (which also would have been drawn in unknown order).

However, I have seen that at least in EMI there is unfortunately a need for having multiple text objects with the same text ID: the entries for the regular menu are created twice: once with color red and a certain layer and once with color black and another layer (but both with the same text ID).

To support this, my proposal was to allow multiple text objects for the same text ID if the layer is different.

Sure, the is some risk:
If the LUA code would generate two text objects with the same text ID and keep references for both and would delete one with KillTextObject, the 2nd reference would be invalid.

Probably it would be better to refcount these objects...

Otherwise I don't know how to solve the original problem: the LUA code "updates" the color of the text objects (for the credits) by just creating new ones, but it doesn't keep any references. Just the fact that the object exists in the engine will cause it to be drawn. To avoid that, either the creation of these many objects must be avoided or the order of creation could have some impact on the order how they are drawn (like their layer).

I would prefer my approach, probably with additional ref-counting. Would that be acceptable?

chkr-private:
After some more investigation I came up with a much better solution which is way less invasive. In favour of the new PR #1193 I'll close this PR without merging.

PR #1191, opened 2015-05-05
WIP: Multisampling for framebuffers

Botje:
I wrote code to enable multisampling for fullscreen framebuffers.
Is there interest in merging this?
I only tested with the shader renderer for now.
This is an OpenGL-only feature, to restore anti-aliasing for people who force it on in their drivers.

Before: https://www.dropbox.com/s/6933731ffq4scu7/without_multisampling.png?dl=0
After: https://www.dropbox.com/s/3iwsxt8kw9gpg5a/with_multisampling.png?dl=0

Known issues:
- [x] Probably linking issues on Windows
- [x] No fallback for Amiga
- [ ] No fallback or error checking in general.
- [x] ADDED: `GfxOpenGLShaders::storeDisplay` is broken (cannot `glReadPixels` from a MS framebuffer

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -458,7 +458,11 @@ Graphics::PixelBuffer SurfaceSdlGraphicsManager::setupScreen(uint screenW, uint 
 			&& !g_engine->hasFeature(Engine::kSupportsArbitraryResolutions)
 			&& framebufferSupported) {
 		glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
-		_frameBuffer = new Graphics::FrameBuffer(fbW, fbH);
+		if (_antialiasing) {
+			_frameBuffer = new Graphics::MultiSampleFrameBuffer(fbW, fbH);
+		} else {
+			_frameBuffer = new Graphics::FrameBuffer(fbW, fbH);
+		}
 		_frameBuffer->attach();
 	}
 #endif
--- a/graphics/opengl/framebuffer.cpp
+++ b/graphics/opengl/framebuffer.cpp
@@ -38,7 +38,9 @@
 #define GL_STENCIL_ATTACHMENT GL_STENCIL_ATTACHMENT_EXT
 #define GL_STENCIL_INDEX8 GL_STENCIL_INDEX8_EXT
 #define GL_DEPTH24_STENCIL8 0x88F0
-#endif
+#define GL_READ_FRAMEBUFFER 0x8CA8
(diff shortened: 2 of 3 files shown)


graphics/opengles2/multisampleframebuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+	glDeleteRenderbuffers(1, &_msColorId);
+	glDeleteRenderbuffers(1, &_msDepthId);
+	glBindFramebuffer(GL_FRAMEBUFFER, 0);
+	glDeleteFramebuffers(1, &_msFrameBufferId);
+}
+
+//TODO: fallback to fewer samples
+void MultiSampleFrameBuffer::init() {
+	glGenFramebuffers(1, &_msFrameBufferId);
+  glBindFramebuffer(GL_FRAMEBUFFER, _msFrameBufferId);
```
klusark: Indentation is a bit off.

bgK:
Sounds like a good idea to have this as a replacement to the "antialiasing" configuration key when rendering to framebuffers.

Botje:
So my frustration with GL extensions had the side effect of using glew first, when possible.
If we add GLEW as a build dependency for OpenGL we can get rid of all that nasty pointer grabbing code , but that _does_ require a glew port for AmigaOS or some more `#ifdef`ing.

bgK:
Requiering GLEW as a dependency for "Advanced" OpenGL support sounds sensible to me. Anyway AmigaOS probably does not support any of these exensions, and would not need the extra functions.

Botje:
There are quite a few problems with this approach, sadly.
One big issue with the multisample framebuffer is that you cannot immediately `glReadPixels` from it without transferring the contents to a regular texture-backed framebuffer.
Quite a few functions take image data from the framebuffer and re-render it later, which means unbind-copy-upload-rebind in several places, adding quite a bit of coupling to the renderer and any framebuffer present.
One way out is using framebuffer data from a previous frame, but that will cause artifacts in some places.
Maybe we should just render in an increased resolution instead, that seems like a far easier approach.

Any thoughts on mainlining the GLEW commits? It doesn't have to be a dependency, but making use of it when present does make for easier maintenance.

aquadran:
GLEW commits looks ok

Botje:
I drastically re-did the implementation, now it reads from the framebuffer used in the last frame.
I checked the uses of `glReadPixels` in our code, and it is only used when taking screenshots and dimming the screen, both of which are not required to be frame perfect IMO.
Now all that is left is a fallback in case the extensions needed are not possible (a small refactor, I think), and this should be good to add.

Any opinions on making antialiasing default to true with this?

bgK:
Having FSAA enabled has no visible effect with Myst III, but shouldn't hurt performance too much either. I don't have strong opinions about having anti-aliasing enabled by default.

bgK:
Superseeded by #1250.

Issue #1046, opened 2014-08-18
EMI: Cannonball is drawn on the ground before the cannon is fired in kab


Akz-:
In the set kab, the cannonball is visible on the ground even before Casaba fires the cannon.

![residualvm 2014-08-18 15-41-38-00](https://cloud.githubusercontent.com/assets/757718/3951379/61a03f3c-26d5-11e4-94ea-7874e1bbd6ff.png)

monkeymadness:
Cannonball still sits on the ground prior to the shot. New cannonball replaces the graphics glitch.

![twocannonballs](https://cloud.githubusercontent.com/assets/10478990/5693231/64115bf2-98d1-11e4-91b3-cb6efc18c87b.jpg)


Version of the game: English version from a jewelcase
OS: Windows 7 64bit
Graphics Card: Intel HD Graphics Driver
Renderer: OpenGL
Version of ResidualVM: 0.3.0gitae24374

chkr-private:
Fixed by merging pull request #1192.

PR #1192, opened 2015-05-05
Implement drawing of dynamically created layers

chkr-private:
This patch set enables the drawing of layers which are dynamically created by EMI's lua code.

* enables the illumination of the neon signs of Planet Threepwood on Jambalaya Island
* turns on water effects in the lagoon on Lucre Island
* hides the cannonball on the beach of Knuttin Atoll until it is actually landed (issue #1046)

### DIFF
--- a/engines/grim/emi/emi.cpp
+++ b/engines/grim/emi/emi.cpp
@@ -143,25 +143,48 @@ void EMIEngine::drawNormalMode() {
 	// Draw actors
 	buildActiveActorsList();
 	sortActiveActorsList();
+	sortLayers();
 
 	Bitmap *background = _currSet->getCurrSetup()->_bkgndBm;
 	background->_data->load();
 	uint32 numLayers = background->_data->_numLayers;
+
+	Common::List<Layer *>::const_iterator nextLayer = _layers.begin();
+	Common::List<Actor *>::const_iterator nextActor = _activeActors.begin();
 	int32 currentLayer = numLayers - 1;
-	foreach (Actor *a, _activeActors) {
-		int sortorder = a->getEffectiveSortOrder();
-		if (sortorder < 0)
-			break;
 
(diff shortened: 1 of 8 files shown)


Botje:
Code-wise this looks good. I don't have access to the game data at the moment so I cannot test, but :+1: from me.

engines/grim/emi/emi.cpp
```diff
@@ -143,25 +143,48 @@ void EMIEngine::drawNormalMode() {
...
+	int bgso = currentLayer * 10;
+
+	// interleave actors, background layers and additional stand-alone layers based
+	// on their sortorder
+	//
+	// priority for same sort order:
+	//   background layers (highest priority)
+	//   stand-alone layers
+	//   actors
+	while (1) {
```
klusark: I feel like this block could be refactored to be much nicer, but I can't think of how. It just feels weird not having a loop condition.

klusark:
Looks good to me. :+1:

engines/grim/emi/emi.cpp
Botje: Something like `while (MAX3(aso, lso, bgso) >= 0) { ... }` maybe?

chkr-private: @klusark, @Botje: I also don't know a nicer way to refactor that. Basically we need  a 3-way-merge (with different kind of objects in each list).

I can factor out the 3 first comparisons <pre>aso >= 0 ...
bgso >= 0 ...
lso >= 0 ...</pre> and use your proposal <pre>while (MAX3(aso, lso, bgso) >= 0) { ... }</pre> From an algorithmic point of view it would be identically. This solution has the benefit that the loop has a defined end condition. My solution on the other hand emphasizes the fact that we would never draw any objects which have a negative sort order. I'm fine with both approaches.

Botje: It is clear enough as is.

PR #1193, opened 2015-05-09
EMI: Correct draw order of textobjects

chkr-private:
Instead of just ordering the textobjects by their layer, use their
creation time (indirectly provided by the pool ID which is continuously
counting up).

This ensures that if there are multiple textobjects with the same text
ID and the same layer, that the latest is drawn last and overwrites the
previous ones.

That's necessary since the LUA code will create multiple instances of
the same textobject for the color fading effect in the credits.

This fixes the issue #1155.


Botje:
:+1: Much simpler :)

PR #1194, opened 2015-05-12
EMI: Keep time of chores and animations in sync

### LINKED ISSUE #1048: EMI: Animation snaps to wrong frame when jumping out of the bank window
Although Guybrush is already on the ground, the animation will snap briefly into a mid-air position.

residualvm 2014-08-18 15-56-20-02

Tested on 1a07a2329.

chkr-private:
Both the animation components/tracks and the chores itself keep track of
the current time independently:

animation component:
- AnimationStateEmi::_time
- initialized with 0
- after the first update() call, _time is advanced by the frametime

chores:
- Costume::_currTime
- initialized with -1
- during the first update() call, _currTime is set first to 0
- during the next update() call, _currTime is advanced by the frametime

EMICostume::update() updates first the chore and then all components
with the same frame time. This has the effect, that the internal time
base of the chore and the animation has always an initial offset (the
value of the frametime of the first update() call).

For some costumes, the chore will pause the animation by setting the
according key for that animation at the end of the animation. However,
since the animation component has a different time base, it has already
reached that point and may have already disabled itself entirely
(instead of just being paused by the chore).

This bug fix avoids setting _currTime to 0 in the first update() call of
the chore but just sets it to the frametime. Any consecutive update()
call will advance the time by the frametime. This will keep the two
time bases of the chore and the animation component in sync.

This fixes #1048.

### DIFF
--- a/engines/grim/emi/animationemi.cpp
+++ b/engines/grim/emi/animationemi.cpp
@@ -100,7 +100,7 @@ Bone::~Bone() {
 
 AnimationStateEmi::AnimationStateEmi(const Common::String &anim) :
 		_skel(nullptr), _looping(false), _active(false), _paused(false),
-		_fadeMode(Animation::None), _fade(1.0f), _fadeLength(0), _time(0), _startFade(1.0f),
+		_fadeMode(Animation::None), _fade(1.0f), _fadeLength(0), _time(-1), _startFade(1.0f),
 		_boneJoints(nullptr) {
 	_anim = g_resourceloader->getAnimationEmi(anim);
 	if (_anim)
@@ -138,7 +138,7 @@ void AnimationStateEmi::update(uint time) {
 	}
 
 	if (!_paused) {
-		uint durationMs = (uint)_anim->_duration;
+		int durationMs = (int)_anim->_duration;
 		if (_time >= durationMs) {
 			if (_looping) {
 				_time = _time % durationMs;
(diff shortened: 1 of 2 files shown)


Akz-:
I suppose another option would be to change AnimationStateEmi so the time would be 0 during the first update after the animation is started? That would ensure the first animation keyframe is always displayed.

chkr-private:
Updated the commit and implemented the bug fix in AnimationStateEmi instead. There is no visible difference. However, probably it is safer to change the animations than the chores.

This PR would change am unsigned value in the savegames to a signed value. So newer savegames would allow negative numbers. Loading a newer savegame with an older engine may cause some trouble, but that can't be prevented. On the other hand, the newer engine can deal with older savegames unless the stored time would be larger than 2 billions. But that's unlikely.

engines/grim/emi/animationemi.cpp
```diff
@@ -138,7 +138,7 @@ void AnimationStateEmi::update(uint time) {
 	}
 
 	if (!_paused) {
-		uint durationMs = (uint)_anim->_duration;
+		int durationMs = (uint)_anim->_duration;
```
Akz-: Unnecessary cast here.

Akz-:
Looks good to me :+1:

chkr-private:
I just updated to commit and corrected the cast.

Issue #1195, opened 2015-05-13
EMI: Game crash when Guybrush finds the secret hideout


monkeymadness:
Testing the game, I came across a new issue that didn't exist before. ResidualVM crashes right as Guybrush finds the secret hideout on Lucre Island.

![game crash](https://cloud.githubusercontent.com/assets/10478990/7602948/8f2533e6-f8e8-11e4-8642-04d2eb37666c.jpg)

Save:
![efmi009](https://cloud.githubusercontent.com/assets/10478990/7602962/bae015e6-f8e8-11e4-86f9-42a0df987264.png)

Version of the game: English version from a jewelcase
OS: Windows 7 64bit
Graphics Card: Intel HD Graphics Driver
Renderer: OpenGL
Version of ResidualVM: 0.3.0git52010ff

chkr-private:
Fixed by commit b186c21f852756e3cdf2809557a6cbb84611d300.

PR #1197, opened 2015-05-16
EMI: Fix wrap around of Layers/Frames

chkr-private:
This patch makes sure, that Layer::_frame is smaller than numLayers
of the bitmap in order to prevent out-of-bounds access to
BitmapData::_layers.

This fixes a crash when finding the secret hideout: #1195.


Botje:
:+1:

PR #1205, opened 2015-08-09
MATH: Fix Line3d::intersectLine2d to do a line segment intersection test

bgK:
It was previously doing a line vs line segment intersection test

I'm a bit confused with this class. It is named Line3d, but actually is a line segment.
I'm also unsure about that change, perhaps the line vs line segment intersection was what was intended.

I tested Grim a bit without noticing any regression.

### DIFF
--- a/math/line3d.cpp
+++ b/math/line3d.cpp
@@ -51,19 +51,25 @@ Math::Vector3d Line3d::middle() const {
 
 bool Line3d::intersectLine2d(const Line3d &other, Math::Vector3d *pos, bool useXZ) {
 
-	float denom, nume_a;
+	float denom, nume_a, nume_b;
 	if (useXZ) {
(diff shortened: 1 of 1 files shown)


aquadran:
I think EMI should be checked too.

Botje:
The call in `actor.cpp` is safe, as the `l` variable is a point in the other sector.
`l` is thus guaranteed to cross the bridge, and the extra check is harmless.

Botje:
As for the calls in `sector.cpp`: the checks on `b1_out` and `b2_out` verify that resp. `bridge.begin` and `bridge.end` lie in the current sector, and the other point is outside. The intersection point must thus also be in between those two points. This call is also safe.

Botje:
:+1: for merging, in other words.

bgK:
EMI seems fine as well.

Thanks for your reviews.

PR #1215, opened 2015-10-09
Fix erroneous access to _aspectCheckbox from in-game options dialog.

gordonacocella:


aquadran:
I'll look into

somaen:
This should presumably be applied upstream in ScummVM then.

bgK:
Merged manually, thanks.

PR #1220, opened 2016-01-01
Move computations out of renderer

vpelletier:
Low-hanging fruits I found while working on lighting last year, where computation could be obviously moved to less-called code.
I do not remember formally benchmarking the difference. Feel free to reject.


Botje:
:+1:

PR #1221, opened 2016-01-01
GRIM: Adds support for loading embedded colormaps.

vpelletier:
Implementing some understanding I had of material file structure while working no lighting code last year.

Very few material files define their own color maps:
  DATA000.LAB/blue.mat
  DATA000.LAB/orange.mat
  DATA000.LAB/pink.mat
  DATA000.LAB/white.mat
  DATA001.LAB/bren_chest.mat
  DATA001.LAB/bren_shin.mat
  DATA001.LAB/bren_skin.mat
  DATA001.LAB/bren_thigh.mat
  DATA004.LAB/bren_chest.mat
  DATA004.LAB/bren_shin.mat
  DATA004.LAB/bren_skin.mat
  DATA004.LAB/bren_thigh.mat
Also, walk (but ignore) downscaled mipmaps. Even fewer files contain them, and
they contain a single image each (with 3 mipmaps):
  DATA000.LAB/dflt.mat
  DATA000.LAB/shadow.mat

### DIFF
--- a/engines/grim/material.cpp
+++ b/engines/grim/material.cpp
@@ -32,12 +32,15 @@
 #include "engines/grim/resource.h"
 #include "engines/grim/textsplit.h"
 
+#define GRIM_MAT_FLAG_HAS_COLORMAP 1
+#define GRIM_MAT_FLAG_UNKNOWN 2
+
 namespace Grim {
 
 Common::List<MaterialData *> *MaterialData::_materials = nullptr;
 
 MaterialData::MaterialData(const Common::String &filename, Common::SeekableReadStream *data, CMap *cmap) :
-		_fname(filename), _cmap(cmap), _refCount(1), _textures(nullptr) {
+		_fname(filename), _cmap(cmap), _refCount(1), _textures(nullptr), _ownCMap(false) {
 
 	if (g_grim->getGameType() == GType_MONKEY4) {
 		initEMI(data);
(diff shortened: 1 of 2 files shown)


aquadran:
How much of testing of Grim was done?

vpelletier:
I played the cutscene where manny meets brennis for the first time, and compared brennis colors with screen-shots found by googling.
Actually, i believe the embedded colormap for bren_* materials is the same as the default one.

This changes was more an effect of taking a longer look at material files than to fix a specific bug (I kind of hoped I would get a hint on why brennis' left jaw is misaligned, but did not reach this).

vpelletier:
I reworked this patch set (...and forgot to squash-or-drop that last commit, I'm working on this but pushed too early):
- Preserve original behaviour of skipping zero-size textures
- Do not fail on 0-mipmaps (aka, no textures)
- Use consistent logging API
- Preserve original order of "set properties" then "check for reasons to skip" so `_data` is `nullptr` when such skip occurs

vpelletier:
And fixed, hopefully properly (man, I fail hard at C++).

engines/grim/material.cpp
```diff
@@ -47,43 +50,72 @@ MaterialData::MaterialData(const Common::String &filename, Common::SeekableReadS
...
 		data->read(t->_data, t->_width * t->_height);
+		/* XXX: Discarding lower-res mipmaps */
+		while (--mipmap_count) {
+			width >>= 1;
+			height >>= 1;
+			data->seek(width * height, SEEK_CUR);
+		}
 	}
+	/* If flags & GRIM_MAT_FLAG_HAS_COLORMAP there is a color map here. */
+        if (flags & GRIM_MAT_FLAG_HAS_COLORMAP) {
```
aquadran: something wrong with formatting here

vpelletier: Oh, nice catch. Fix pushed (and rebased on current master).

PR #1225, opened 2016-01-03
STARK: adds support for tooltips in inventory

dpentz:
Adds support for tooltips in inventory screen for TLJ


dpentz:
The latest commit handles that as I've seen in the original game.

bgK:
There is now an alternate implementation in the main repo.

PR #1226, opened 2016-01-05
WIN32: Embed the shaders in the executable

bgK:
Our users are used to ResidualVM's executable not to require extra files, as can be seen there http://forums.residualvm.org/viewtopic.php?f=1&t=712. This embeds the shaders files in the Windows executable.

I'm not happy with having to list all the shaders in the .rc file, but can't think of a better solution. Putting the shaders in some kind of archive would be nice. But I don't know if it can be done easily when building with Visual Studio.

This was not tested with Visual Studio.

### DIFF
--- a/dists/residualvm.rc
+++ b/dists/residualvm.rc
@@ -18,6 +18,47 @@ modern.zip        FILE    "gui/themes/modern.zip"
 translations.dat       FILE    "gui/themes/translations.dat"
 #endif
 
+#ifdef USE_OPENGL_SHADERS
+#if ENABLE_GRIM == STATIC_PLUGIN
+shaders/dim.fragment             FILE    "engines/grim/shaders/dim.fragment"
+shaders/dim.vertex               FILE    "engines/grim/shaders/dim.vertex"
+shaders/emerg.fragment           FILE    "engines/grim/shaders/emerg.fragment"
+shaders/emerg.vertex             FILE    "engines/grim/shaders/emerg.vertex"
+shaders/emi_actor.fragment       FILE    "engines/grim/shaders/emi_actor.fragment"
+shaders/emi_actor.vertex         FILE    "engines/grim/shaders/emi_actor.vertex"
+shaders/emi_background.fragment  FILE    "engines/grim/shaders/emi_background.fragment"
+shaders/emi_background.vertex    FILE    "engines/grim/shaders/emi_background.vertex"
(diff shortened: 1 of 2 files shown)


somaen:
There is precedence for requiring extra files (gfupd101.exe in game data folder), as well as for us to supply files that need to be available (see dists/engine-data). So not entirely sure about this.

bgK:
@somaen the files in dists/engine-data are already embedded in the executable.

aquadran:
I don't know why we have them in the executable. Only reason I see is when you develop with MSVC to able load them. For that for unix platforms we have hack by adding paths in gles shader code. maybe we can add relative path for visual studio too.

aquadran:
btw, we don't even have detection which engine is enabled.

bgK:
@aquadran We have the engine data files in the executable in order to ensure their version is synced with the executable. Also so that the users don't have to care about them / update them. ScummVM does the same. I'd very much like to do the same with the shaders. Not having to deal with bugs due to mismatched shaders version seems like a good enough reason to me.
Also neither the engine data files nor the shaders are shipped in the windows installer.

aquadran:
@bgK Other platforms can have the same issue, windows is not exception here. Windows executable can load data from files like others ports. We can simply copy files into executable (sub)directory. If you worry about mismatched versions, we could add marker (residualvm version) to check at runtime.

bgK:
@aquadran Yes, we can copy them in a subdirectory, but why introduce a new way of doing things when all the other files are embedded in the executable? I mean the engine data files, the theme and the translations are already in the executable.
The grim patches mentioned by somaen arguably can't be shipped with ResidualVM due to licence reasons.
All of our supported platforms are specific ways of dealing with those files. Android embeds them in the APK, OSX puts them in the app bundle, Debian copies them in /usr/share/residualvm, ... These ways solve the missing files and mismatched versions issue efficiently. The win32 way we use as well.

aquadran:
@bgK My point is not where files are copied. All ports do own way as windows. What exception is here only windows resource file define what files to include. None other platforms do that. This is why I don't like idea to put such things here.

Kirben:
Adding the shaders to the ResidualVM executable is definitely the best option, especially as they are very small files. It means the ResidualVM executable is more portable, and users don't have to worry about any essential files been missing

If the separate shaders were added to the Windows installer script, I would have to assume that all the shaders are going to be bundled, no matter the compile options used.

aquadran:
For long term it would be better use archive like zip. For short, ok, let's merge this.

PR #1236, opened 2016-02-06
GRAPHICS: Check OpenGL context properties at runtime

bgK:
Changes summary:
- Add an OpenGL::Context singleton to query at runtime the current OpenGL context capabilities
- Add an OpenGL::SurfaceRenderer class to abstract drawing textures as 2D surfaces
- Make use of these features to clean up the SurfaceSdlGraphicsManager class
- Make use of these features to gracefully fallback to an available game renderer when the current OpenGL context does not permit using the user selected renderer
- Use GLSL 1.20 for desktop GL contexts on all platforms (previously MacOS used 1.20, the other platforms 1.30)

This could also be used as a foundation to select GLES2 contexts at runtime on desktop platforms once we support SDL2.


aquadran:
:+1:

bgK:
@Botje, @somaen do you want to look at this before it is merged?

Botje:
I just ran through all your commit. Looks good! :shipit:

PR #1238, opened 2016-03-06
SDL: Reorganize the game render target selection code

bgK:
Introduce a GameRenderTarget enumeration defining how the game is rendered
on the screen: directly, using a sub-surface, or using a framebuffer.

Split the video initialisation code into simpler methods.
The logic should be exactly the same as before.

The underlying goal is to hopefully make it easier to port that portion to SDL2 in the future.


aquadran:
:+1:

Botje:
:+1:

PR #1241, opened 2016-04-22
STARK: Add support for German TLJ 4CD version

laenion:
These are the hashes of the original German retail version.


bgK:
Thanks

PR #1250, opened 2016-06-04
Multisampling for framebuffers

bgK:
The Grim engine draws to a framebuffer, when used in fullscreen mode, in order to use the desktop resolution and scale keeping a correct aspect ratio. This is an updated version of @Botje's work on enabling anti-aliasing when rendering to framebuffers (#1191).

I have added error checking and fallbacks to Botje's work, so that it should be in a mergable state.


bgK:
@klusark: ranlib: engines/grim/libgrim.a: No space left on device :wink:

aquadran:
👍

PR #1252, opened 2016-06-13
Basic port to SDL2

bgK:
This PR contains minimal changes to make ResidualVM work with SDL2. It is based on previous work by @aquadran and the ScummVM team.

Features are the same as with SDL1. The only exception being that it is now possible to use GLES2 contexts when building with SDL2.

Configuring with SDL2 is done using the following command:
`SDL_CONFIG=sdl2-config ./configure`

Configuring with SDL2 for GLES2 is done using the following command:
`SDL_CONFIG=sdl2-config ./configure --force-opengles2 --disable-glew --enable-opengl-shaders`
This command looks like a black magic invocation, and should probably be simplified at a later point.

I've done tests without noticing bugs in the following configurations:
- Linux OpenGL (mesa/radeonsi)
- Linux GLES2 (mesa/radeonsi)
- Windows 7 OpenGL (AMD Catalyst)
- Windows 7 GLES2 (ANGLE) (Manny is not visible)

I'd be curious to have test results with OS X, Raspberry PI, (and AmigaOS, if the SDL2 port is mature enough).

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -57,6 +57,7 @@ SurfaceSdlGraphicsManager::SurfaceSdlGraphicsManager(SdlEventSource *sdlEventSou
 	SdlGraphicsManager(sdlEventSource, window),
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	_renderer(nullptr), _screenTexture(nullptr),
+	_glContext(nullptr),
 #endif
 	_screen(0),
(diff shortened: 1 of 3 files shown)


raziel-:
SDL2 might work on AmigaOS(4), but GLES is not (yet) available

bgK:
Any feedback?

aquadran:
sorry, I'll check this today on osx. Regarding code, only thing it's compilation on os x, but I guess this need to be fixed in scummvm side too.

aquadran:
👍

PR #1256, opened 2016-07-01
SDL: Split the OpenGL code to a separate GraphicsManager

bgK:
The general goal for these changes is to move the OpenGL specific code out of SurfaceSdlGraphicsManager so that the code is less messy, and it is easier to add features. Possibly in the future the OpenGL code could be further separated from the SDL code so that it can be reused for other platforms.

A new class, ResVMSdlGraphicsManager has been added in the class hierarchy between SurfaceSdlGraphicsManager and SdlGraphicsManager to be able to share common code between the surface and OpenGL graphics managers. The OpenGL code has been moved to OpenGLSdlGraphicsManager, subclassing ResVMSdlGraphicsManager.

The SDL platform code now chooses to instantiate either SurfaceSdlGraphicsManager or OpenGLSdlGraphicsManager based on the "accel3d" parameter of the "setupScreen" method.

The OpenGL code used to draw the overlay has been extracted to a new class named "TiledSurface". This new class has slightly better dirty texture handling which could improve performance a bit.

The way the overlay is drawn when in OpenGL mode was changed. There are now two 2D surfaces being rendered instead of one. The first one contains a capture of the game screen. The second one is drawn on top of it with alpha blending enabled and contains the actual overlay / GMM. This finally fixes the glitch where GUI dialogs would not disappear when closed in the GMM, at the cost of additional VRAM usage.

### DIFF
--- a/backends/graphics/sdl/resvm-sdl-graphics.cpp
+++ b/backends/graphics/sdl/resvm-sdl-graphics.cpp
@@ -1,0 +1,367 @@
+/* ResidualVM - A 3D game interpreter
+ *
+ * ResidualVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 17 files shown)


Botje:
Nice refactor! 👍

bgK:
I'll merge this in a few days unless somebody wants to further review it.

aquadran:
I want review it :)

backends/graphics/sdl/resvm-sdl-graphics.cpp
```diff
@@ -0,0 +1,321 @@
...
+static const OSystem::GraphicsMode s_supportedGraphicsModes[] = {
+		{0, 0, 0}
+};
+
+ResVmSdlGraphicsManager::ResVmSdlGraphicsManager(SdlEventSource *source, SdlWindow *window, const Capabilities &capabilities) :
+		SdlGraphicsManager(source, window),
+		_fullscreen(false),
+		_lockAspectRatio(true),
+		_overlayVisible(false),
+        _overlayWidth(0),
```
aquadran: formatting

aquadran:
:+1:

backends/graphics/sdl/resvm-sdl-graphics.cpp
bgK: fixed

Issue #1186, opened 2015-04-05
GRIM: Lighting is enabled for roulette wheels in opengl + shaders mode.


vpelletier:
Likely related to #264 fix being somehow missing in shader renderer.

vpelletier:
Fixed by #1258 .

PR #1258, opened 2016-07-09
GRIM, SHADERS: Toggle light inside drawMesh.

vpelletier:
Fixes appearance of roulettes in cn and beavers in dm (#1186).


bgK:
:+1:

vpelletier:
Reworked patch 1 to not enable lighting in shadow mode.
Added patch 2 to remove dead calls to *ableLight inside the loop in `drawMesh`.
Added patch 3 to reduce differences between GRIM and EMI actor vertex shaders. The main purpose here is to make obvious that `lightsEnabled` and `shadow._active` are not necessarily mutually exclusive (as far as shaders are concerned).

aquadran:
:+1:

PR #1259, opened 2016-07-10
(DO NOT MERGE) TINYGL: Ignore zero-coordinate-delta textures.

vpelletier:
The holepunch in ha has what is likely a data error: all textures
coordinates are (0, 0). TinyGL used to render the holepunch in the (likely
intended) dark blue color from the texture. But this is inconsistently
handled between TinyGL and OpenGL (with and without shaders): the latter
ignore the texture, and render the holepunch as shaded grey. So make
TinyGL behave consistently - data error can be fixed independently.

Original Grim software renderer display the holepunch in a very dark grey
color, making it hard to tell if it takes the texture into account at all.
Data remained the same in remastered edition, which displays the holepunch
as light grey (on linux at least), consistently with residual OpenGL
renderer.

Note: this modifies bug #58 as now all renderers are consistent. But the
data error remains.

### DIFF
--- a/graphics/tinygl/clip.cpp
+++ b/graphics/tinygl/clip.cpp
@@ -409,6 +409,7 @@ void gl_draw_triangle_fill(GLContext *c, GLVertex *p0, GLVertex *p1, GLVertex *p
 		count_triangles++;
 	}
 #endif
+	const Vector4 default_tex_coord = Vector4(0.0f, 0.0f, 0.0f, 1.0f);
 
 	if (c->color_mask == 0) {
 		// FIXME: Accept more than just 0 or 1.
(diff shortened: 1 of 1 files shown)


vpelletier:
This causes a regression in the black reaper costume (set lr), several triangles appear in a shaded white. A few more appear black when replacing the "&&" between coordinates into an "||" (ie, texture is ignored only if all texture vertices are in the same spot) but several white triangles remain.

vpelletier:
And I found a counter-example: Glottis' fingers have texture coordinates (0, 0). Still, OpenGL do set a color on them (which should AFAIK come from the texture).
So this patch causes more regressions (8 fingers, at least) than it fixes (one holepunch) :) .
Should this pull-request be closed (without merge, of course) ?

aquadran:
@vpelletier maybe just add to title as "Will not merge" or something like that.

PR #1260, opened 2016-07-10
GRIM, TINYGL: Always check if texture matrix need applying.

vpelletier:
matrix_model_projection_updated is not set when texture matrix alone is
modified.
Allows getting rid of a few TinyGL-only lines.


vpelletier:
There is something wrong in this pull-request, likely when I rebased it on master (I have been poking around in tinygl too much). Withdrawing for now.

PR #1262, opened 2016-07-11
GRIM: Fix drawLine/Rect/Polygon in TinyGL renderer

### LINKED ISSUE #1181: GRIM: Scroll arrows when loading game not visible with TinyGL
Current Git snapshot, English version of the game.

When loading a game (at least if there are enough savegames so that they aren't all visible at once), there are supposed to be triangular arrows indicating that the list continues. These work for me with the OpenGL renderer (unless I enable shaders), but not with the TinyGL renderer.

It looks like the arrows are made up of filled rectangles. It does call drawRectangle(), and the parameters look sensible enough, but for some reason they just don't show up. Maybe the problem is with BlastRect(), rather than drawRectangle()?

vpelletier:
Two short changes to TinyGL which allow copying OpenGL renderer implementation of these methods over to TinyGL one.

Fixes #1181 .

### DIFF
--- a/engines/grim/gfx_opengl.cpp
+++ b/engines/grim/gfx_opengl.cpp
@@ -1957,33 +1957,12 @@ void GfxOpenGL::drawRectangle(const PrimitiveObject *primitive) {
 		glVertex2f(x1, y2 + 1);
 		glEnd();
 	} else {
-		glBegin(GL_QUADS);
-
-		// top line
+		glLineWidth(_scaleW);
+		glBegin(GL_LINE_LOOP);
 		glVertex2f(x1, y1);
 		glVertex2f(x2 + 1, y1);
-		glVertex2f(x2 + 1, y1 + 1);
-		glVertex2f(x1, y1 + 1);
-
-
-		// right line
-		glVertex2f(x2, y1);
-		glVertex2f(x2 + 1, y1);
-		glVertex2f(x2 + 1, y2 + 1);
-		glVertex2f(x2, y2);
-
-		// bottom line
-		glVertex2f(x1, y2);
-		glVertex2f(x2 + 1, y2);
 		glVertex2f(x2 + 1, y2 + 1);
 		glVertex2f(x1, y2 + 1);
-
-		// left line
-		glVertex2f(x1, y1);
-		glVertex2f(x1 + 1, y1);
-		glVertex2f(x1 + 1, y2 + 1);
-		glVertex2f(x1, y2);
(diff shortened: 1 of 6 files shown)


vpelletier:
For (t)glOrtho, here is (the documentation)[https://www.opengl.org/sdk/docs/man2/xhtml/glOrtho.xml] . The page uses mathml, so chromium won't display the most interesting part - use firefox.

vpelletier:
With these 2 latest commits, Copal's computer is properly displayed.
I also have a fix for the ticket printer in vpelletier/residualvm@59f00d49ad070f995aadd51c04a1a6a4d973ea38 but I am not sure whether I should put that commit in this pull-request too.

I believe the ticket printer is the last issue (after this pull request goes in) which prevents GRIM from being completable with TinyGL alone.

engines/grim/gfx_opengl.cpp
```diff
@@ -1957,33 +1957,12 @@ void GfxOpenGL::drawRectangle(const PrimitiveObject *primitive) {
 		glVertex2f(x1, y2 + 1);
 		glEnd();
 	} else {
-		glBegin(GL_QUADS);
-
-		// top line
+		glLineWidth(_scaleW);
```
bgK: Is taking care of the case where _scaleH is not equal to _scaleW is useful?

vpelletier: Good question. I took this call from GfxOpenGL::drawLine just below. Maybe I should add a comment about this approximation ? And should I withdraw the opengl part of my change, as it does not change the rendering, but merely tries to avoid increasing the diff between OpenGL and TinyGL ?

bgK: I'd say it's ok to keep it the way you did if other methods do the same approximation.
To my understanding, _scaleW and _scaleH are not equal to one when the screen size is not the same as the game size which is not possible currently (but would be very nice). And even then, _scaleW and _scaleH would have different values only when the aspect ratio correction is disabled.

bgK:
:+1:

bgK:
Maybe you can do another PR for https://github.com/vpelletier/residualvm/commit/59f00d49ad070f995aadd51c04a1a6a4d973ea38, since it is a bit more involved.

vpelletier:
Is there anything blocking this PR ? Note that it does not depend on #1266 . Alone, it will still improve the ticket printer in that it will display something - just not readable digits or letters.

PR #1265, opened 2016-07-14
TinyGL: Fix TGL_ALPHA_TEST support with clipped triangles

vpelletier:
Before the first patch, enabling TGL_ALPHA_TEST in GRIM would cause clipped faces to completely disappear (with a striking effect in set mo when near the tube): their vertices' color would become 0 because of updateTmp saving & restoring W color component while it was never set before. And of course, enabling alpha testing makes these vertices go away along with their textures.
The TinyGL patch corrects this original issue, and the GRIM patch makes use of alpha test.

As a result, alpha textures (like the card) look better than before.

### DIFF
--- a/graphics/tinygl/clip.cpp
+++ b/graphics/tinygl/clip.cpp
@@ -226,13 +226,12 @@ float(*clip_proc[6])(Vector4 *, Vector4 *, Vector4 *) =  {
 
 static inline void updateTmp(GLContext *c, GLVertex *q, GLVertex *p0, GLVertex *p1, float t) {
 	if (c->current_shade_model == TGL_SMOOTH) {
-		float a = q->color.W;
 		q->color = p0->color + (p1->color - p0->color) * t;
-		q->color.W = a;
 	} else {
 		q->color.X = (p0->color.X);
 		q->color.Y = (p0->color.Y);
 		q->color.Z = (p0->color.Z);
+		q->color.W = (p0->color.W);
 	}
 
 	if (c->texture_2d_enabled) {
--- a/engines/grim/gfx_tinygl.cpp
+++ b/engines/grim/gfx_tinygl.cpp
@@ -730,9 +730,13 @@ void GfxTinyGL::drawEMIModelFace(const EMIModel *model, const EMIMeshFace *face)
 }
 
 void GfxTinyGL::drawModelFace(const Mesh *mesh, const MeshFace *face) {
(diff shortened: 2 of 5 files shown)


graphics/tinygl/clip.cpp
```diff
@@ -226,13 +226,12 @@ float(*clip_proc[6])(Vector4 *, Vector4 *, Vector4 *) =  {
 
 static inline void updateTmp(GLContext *c, GLVertex *q, GLVertex *p0, GLVertex *p1, float t) {
 	if (c->current_shade_model == TGL_SMOOTH) {
-		float a = q->color.W;
 		q->color = p0->color + (p1->color - p0->color) * t;
-		q->color.W = a;
```
aquadran: what about alpha channel here?

vpelletier: The alpha channel is `color.W`, right ?
`q`, `p0` and `p1` being GLVertex, `->color` is Vector4 and the operators it defines (`Vector4 + Vector4`, `Vector4 - Vector4` and `Vector4 * float`) all update the `W` component.

vpelletier: BTW, doesn't C++ create 2 (or maybe even 3) intermediate Vector4 instances in the
```c++
q->color = p0->color + (p1->color - p0->color) * t;
```
line ? One for `-`, one for `*`, and one for `+`. I do not know if the last one is copied to existing `q->color` Vector4 instance or if it replaces it.

vpelletier:
This extra commit fixes a minor visual difference between OpenGL and TinyGL with the end of the tube (the 3d shape): it is now transparent due to alpha test, but the side facing Manny would not be displayed although back face culling is disabled.

It creates a conflict with #1266 though. Which one should I rebase on top of the other ?

graphics/tinygl/clip.cpp
bgK: The fix looks good to me. There's no reason to do special handling for the alpha component when interpolating.

Out of curiosity, I checked the disassembly for that function on x86_64 with GCC6.1 / O2. No Vector4 temporaries are created at all. Probably thanks to "return value optimization". The generated code is quite optimized, but not as much as I would have hoped. SIMD instructions are not used to perform the operations on all the vector components at once.

BTW, the assignment below can be done as "q->color = p0->color", now.

bgK:
:+1:

aquadran:
:+1:

PR #1266, opened 2016-07-14
TINYGL: Simplify zline.cpp .

vpelletier:
drawLine was broken for dx = 0 lines.
Also, the split between fillLineGeneric and drawLine, along with the many
parameters exchanged (with obscure names with subtle differences), made
the code hard to understand.
This implementation should be easier to check.
There was no visible CPU usage change.
Fixes GRIM's ticket printer (#1262 is needed too), which are composed of a lot of vertical
segments.

I am not used to C++, so there may be possible simplifications. For example:
- I split `putPixel` to avoid a "variable may be used before being initialised" warning, although it should be harmless in practice (the 3rd argument would be used in `putPixel` when `kInterpZ` is false anyway).
- I made `putPixel` and `drawLine` into (private) methods rather that static functions because this seems more natural than passing them a `this` argument. I have no idea if this has an adverse effect (performance ?).

### DIFF
--- a/graphics/tinygl/zbuffer.h
+++ b/graphics/tinygl/zbuffer.h
@@ -411,9 +411,6 @@ struct FrameBuffer {
 	template <bool kInterpRGB, bool kInterpZ, bool kInterpST, bool kInterpSTZ, int kDrawMode>
 	void fillTriangle(ZBufferPoint *p0, ZBufferPoint *p1, ZBufferPoint *p2);
 
-	template <bool kInterpRGB, bool kInterpZ, bool kDepthWrite>
-	void fillLineGeneric(ZBufferPoint *p1, ZBufferPoint *p2, int color);
-
 	void fillTriangleTextureMappingPerspectiveSmooth(ZBufferPoint *p0, ZBufferPoint *p1, ZBufferPoint *p2);
 	void fillTriangleTextureMappingPerspectiveFlat(ZBufferPoint *p0, ZBufferPoint *p1, ZBufferPoint *p2);
 	void fillTriangleDepthOnly(ZBufferPoint *p0, ZBufferPoint *p1, ZBufferPoint *p2);
@@ -425,9 +422,9 @@ struct FrameBuffer {
 	void plot(ZBufferPoint *p);
 	void fillLine(ZBufferPoint *p1, ZBufferPoint *p2);
 	void fillLineZ(ZBufferPoint *p1, ZBufferPoint *p2);
-	void fillLineFlatZ(ZBufferPoint *p1, ZBufferPoint *p2, int color);
+	void fillLineFlatZ(ZBufferPoint *p1, ZBufferPoint *p2);
 	void fillLineInterpZ(ZBufferPoint *p1, ZBufferPoint *p2);
-	void fillLineFlat(ZBufferPoint *p1, ZBufferPoint *p2, int color);
+	void fillLineFlat(ZBufferPoint *p1, ZBufferPoint *p2);
 	void fillLineInterp(ZBufferPoint *p1, ZBufferPoint *p2);
 
 	void setScissorRectangle(int left, int right, int top, int bottom) {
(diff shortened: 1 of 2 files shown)


vpelletier:
Rebased over latest master.

vpelletier:
After this pull request, I have a slight improvement for non-blit hotspots found by "perf top" around scissorPixel. See [vpelletier/residualvm@tinygl_scissorpixel](https://github.com/vpelletier/residualvm/tree/tinygl_scissorpixel).

Besides blitting, another hot spot seems to be loading texture from memory. I picked the first hot spot I could see how to improve, but as a result the gain is quite low.

If desired, I can include these commits in this pull-request.

aquadran:
I'm not sure about this PR. Which part fix described issue?

vpelletier:
I'm not sure if there is a single place to point out which solves the issue. At least, the major part of the change is in resulting `drawLine`, the rest is just following on this. I think I also fix a coordinate bug in `plot`, as the `PSZB` factor would suggest a byte index while `putPixel` expected a pixel index. On a more abstract level I would argue that after this change the code is easier to understand (although I'm obviously totally biased), which can help spotting and fixing issues.

I can point at least at one place in the original code which illustrates the issue with vertical lines: `int n = dx;` in `drawLine`. This can be reached with `dx = 0` because caller only checks for dx and dy being simultaneously zero, then for dx being strictly greater than zero, then handles all other possibilities in the final `else` block by still calling the same function with same parameters. Replacing with (`int n = dx > dy ? dx : dy;`) alone is not enough, because then the lines which should be vertical are now displayed at a 45° angle.

As for what I do not understand in the original code (and I would be just happy to have them described, as I am certainly missing on implementation subtleties):
- why the `fillLineGeneric` vs. `drawLine` separation ? `drawLine` parameter list is long with some hard to understand names (`inc_1`, `inc_2`) and some kind of redundancy (`r`, `g`, `b` vs. `color` vs. color component of `p1` and `p2`) and seem to mix different approaches (`pz` which is a pointer to a "pixel" in z buffer while `buffer` and `pixelOffset` are separate)
- why isn't `drawLine` a method when it receives an instance as first argument ?
- I just do not understand why a, dy and dx are processed the way they are before the main loop in `putPixel` (especially why dx becomes `dx * 2 + dy * 2` - if we ignore the `dy = dy * 2` line). In Bresenham's line drawing algorithm the error (which seems to be `a` here) is half of dx or dy, but here all 3 values are multiplied by 2. Maybe this is a different algorithm ?
- Why is there an 8-bits left-shift of rgb components in `fillLineGeneric` but none in `plot`, while in both case they get passed to RGB_TO_PIXEL (and an example from ztriangle.cpp suggests `plot` way would be the correct one, but I did not check much further than "rendering looks correct") ?

This said, I see now that I expanded two color comparisons into rgb properties comparisons only, which is going the wrong direction.

vpelletier:
> - Why is there an 8-bits left-shift of rgb components [...]

I checked the origin of these values further (gl_transform_to_viewport and tglColor4f), and they are already left-shifted by (at least) 8:
```python
# red & blue extremum
>>> hex(int(1.0 * ((1<<16) - (1<<10) - (1<<10)) + (1<<10)))
'0xfc00'
>>> hex(int(0.0 * ((1<<16) - (1<<10) - (1<<10)) + (1<<10)))
'0x400'
# green extremum
>>> hex(int(1.0 * ((1<<16) - (1<<9) - (1<<9)) + (1<<9)))
'0xfe00'
>>> hex(int(0.0 * ((1<<16) - (1<<9) - (1<<9)) + (1<<9)))
'0x200'
```
BTW, it is not obvious to me why colors wouldn't be kept in the full 0x00..0xFF range (or 0x0000..0xFFFF with lsB being the fractional part) until they get converted to some packed color suiting the framebuffer format: "perfect" black and white is not possible with current code (which may have no visual effect at all).

> This said, I see now that I expanded two color comparisons into rgb properties comparisons only, which is going the wrong direction.

And I somehow thought colors were available as a vector property on ZBufferPoint, while they are stored as separate r/g/b properties. So I think this part is fine.

bgK:
I haven't compared it with the old code, but the new code looks good to me. It is indeed much better, thanks.

vpelletier:
Rebased on master (no other change)

vpelletier:

PR #1267, opened 2016-07-15
GRIM: Enable TGL_ALPHA_TEST in GfxTinyGL::drawSprite.

vpelletier:
Forgotten in #1265 . I do not know where this has a visible effect (smoke from Manny's idle animation seemed already alpha-blended with background without this).


aquadran:
don't forget tests with EMI which use alpha blending a lot

PR #1269, opened 2016-07-16
Tinygl: optimise gl_shade_vertex

vpelletier:
Avoids or deffer normalising when possible.
Avoids computing specular when light & material combination would produce no specular effect.

### DIFF
--- a/graphics/tinygl/light.cpp
+++ b/graphics/tinygl/light.cpp
@@ -58,6 +58,7 @@ void glopMaterial(GLContext *c, GLParam *p) {
 		break;
 	case TGL_SPECULAR:
 		m->specular = v;
+		m->has_specular = v.X != 0 || v.Y != 0 || v.Z != 0 || v.W != 1;
 		break;
 	case TGL_SHININESS:
 		m->shininess = v.X;
@@ -99,6 +100,7 @@ void glopLight(GLContext *c, GLParam *p) {
 		break;
 	case TGL_SPECULAR:
 		l->specular = v;
+		l->has_specular = v.X != 0 || v.Y != 0 || v.Z != 0 || v.W != 1;
 		break;
 	case TGL_POSITION: {
 		Vector4 pos;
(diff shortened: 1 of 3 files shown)


graphics/tinygl/light.cpp
```diff
@@ -237,7 +237,7 @@ void gl_shade_vertex(GLContext *c, GLVertex *v) {
 			d.Y = l->position.Y - v->ec.Y;
 			d.Z = l->position.Z - v->ec.Z;
 			dist = sqrt(d.X * d.X + d.Y * d.Y + d.Z * d.Z);
-			if (dist > 1E-3) {
+			if (dist != 1) {
```
tombsar: Is this case ever hit in practice? Only very particular vectors will have a norm of exactly 1.0f.

tombsar: Also, if you're doing this test, you could also move the expensive `sqrt` inside, as there is no point finding the square root of unity.

graphics/tinygl/light.cpp
```diff
@@ -237,7 +237,7 @@ void gl_shade_vertex(GLContext *c, GLVertex *v) {
 			d.Y = l->position.Y - v->ec.Y;
 			d.Z = l->position.Z - v->ec.Z;
 			dist = sqrt(d.X * d.X + d.Y * d.Y + d.Z * d.Z);
-			if (dist > 1E-3) {
```
tombsar: This test seems to be about avoiding dividing by very small numbers. By re-purposing it for optimization, are you sure that, for example, a divide by zero cannot now occur?

graphics/tinygl/light.cpp
vpelletier: Good points, I will push a modified commit.

graphics/tinygl/light.cpp
vpelletier: Division by zero gets fixed in the last commit, which moves normalisation inside `if (dot > 0)` block. I'll fix it in the intermediate state too.

tombsar: Sorry, missed that the dot product guarantees it indirectly. I withdraw my objection (and probably no need to fix the intermediate state either).

graphics/tinygl/light.cpp
```diff
@@ -228,88 +230,93 @@ void gl_shade_vertex(GLContext *c, GLVertex *v) {
...
+						s.Y = d.Y;
+						s.Z = (float)(d.Z + 1.0);
+					}
+					dot_spec = n.X * s.X + n.Y * s.Y + n.Z * s.Z;
+					if (twoside && dot_spec < 0)
+						dot_spec = -dot_spec;
+					if (dot_spec > 0) {
+						GLSpecBuf *specbuf;
+						int idx;
+						tmp = sqrt(s.X * s.X + s.Y * s.Y + s.Z * s.Z);
```
tombsar: Potential optimization here: depending how it is implemented, it should be cheaper to calculate the inverse square root here, then multiply `dot_spec` below.

tombsar: Also same comment as for `dist` above; no need to do the square root outside of the test for unity.

vpelletier: I am not sure this still applies to latest version.

tombsar: Second comment no longer applies, but I believe there is still an opportunity to avoid the expensive sqrt call. I just don't know how to get the compiler to emit a reciprocal square root without intrinsics...

vpelletier:

PR #1270, opened 2016-07-16
Tinygl 32bits

vpelletier:
TinyGL is now looking nice, with smooth color transitions.

I did not notice a cpu usage increase. Actually, as there are fewer shift/mask operations I would actually expect a (very) tiny speedup.

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -87,7 +87,7 @@ void SurfaceSdlGraphicsManager::setupScreen(uint gameWidth, uint gameHeight, boo
 	if (_fullscreen)
 		sdlflags |= SDL_FULLSCREEN;
 
-	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 16, sdlflags);
+	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 0, sdlflags);
 	if (!_screen) {
 		warning("Error: %s", SDL_GetError());
 		g_system->quit();
@@ -416,13 +416,13 @@ SDL_Surface *SurfaceSdlGraphicsManager::SDL_SetVideoMode(int width, int height, 
 		return nullptr;
 	}
 
-	_screenTexture = SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_RGB565, SDL_TEXTUREACCESS_STREAMING, width, height);
+	_screenTexture = SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, width, height);
 	if (!_screenTexture) {
 		deinitializeRenderer();
 		return nullptr;
 	}
 
-	SDL_Surface *screen = SDL_CreateRGBSurface(0, width, height, 16, 0xF800, 0x7E0, 0x1F, 0);
+	SDL_Surface *screen = SDL_CreateRGBSurface(0, width, height, 32, 0x00FF0000, 0x0000FF00, 0x000000FF, 0xFF000000);
 	if (!screen) {
 		deinitializeRenderer();
 		return nullptr;
--- a/graphics/tinygl/ztriangle.cpp
+++ b/graphics/tinygl/ztriangle.cpp
@@ -34,35 +34,27 @@ namespace TinyGL {
 
 static const int NB_INTERP = 8;
 
-#define SAR_RND_TO_ZERO(v,n) (v / (1 << n))
-
 template <bool kDepthWrite, bool kEnableAlphaTest, bool kEnableScissor, bool kEnableBlending>
 FORCEINLINE static void putPixelFlat(FrameBuffer *buffer, int buf, unsigned int *pz, int _a,
-                                     int x, int y, unsigned int &z, int color, unsigned int &a, int &dzdx) {
+                                     int x, int y, unsigned int &z, unsigned int &r, unsigned int &g, unsigned int &b, unsigned int &a, int &dzdx) {
 	if ((!kEnableScissor || !buffer->scissorPixel(x + _a, y)) && buffer->compareDepth(z, pz[_a])) {
-		unsigned int r = (color & 0xF800) >> 8;
-		unsigned int g = (color & 0x07E0) >> 3;
-		unsigned int b = (color & 0x001F) << 3;
-		buffer->writePixel<kEnableAlphaTest, kEnableBlending, kDepthWrite>(buf + _a, a / 256, r, g, b, z);
+		buffer->writePixel<kEnableAlphaTest, kEnableBlending, kDepthWrite>(buf + _a, a / 256, r / 256, g / 256, b / 256, z);
 	}
 	z += dzdx;
 }
 
 template <bool kDepthWrite, bool kEnableAlphaTest, bool kEnableScissor, bool kEnableBlending>
 FORCEINLINE static void putPixelSmooth(FrameBuffer *buffer, int buf, unsigned int *pz, int _a,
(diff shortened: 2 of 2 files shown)


aquadran:
16bit mode is left for speed instead 32bit mode

vpelletier:
So it does make a performance difference in favor of 16 bits ? I'll profile again.
Note that these change do not force 32 bits mode, as the choice is left to SDL.

aquadran:
please test the game, not profile one part of tinygl code. switching to 32bits it will affect rest of gfx drawing.

vpelletier:
When I say profile, I mean using `perf top` and not enabling the tinygl profiling statistics. `perf top` profiles the whole system, and so residualvm as a whole.

I also played a few sets (ex: cn to have many textured triangles thanks to the 7 actors + 3 spinners) and performance was fine.

Did you notice performance issues with these patches ? In which set ?

aquadran:
do you test grim or emi?

vpelletier:
Updated patch stack: fixes purple rectangles in EMI.
Current patch stack is tested with SDL framebuffer in 16 and 32 bits modes, with EMI (ship scene at game start, there is transparency between ropes in the background) and GRIM (first game scene, as tube image contains transparency).

vpelletier:
I did a quick FPS comparison on the ship scene at the beginning of EMI with all my pending pull-requests. I wait for ~1 minute watching the FPS counter to see the value range, so the measures are approximate. I patched  grim.cpp to disable FPS limit when `engine_speed` is 0. Should I open a PR for that patch (maybe include in #1271 and add a field to change that value too) ? The original code looks like it really wants to have an FPS limit in all cases.

gcc version 5.4.0 20160609 (Debian 5.4.0-5)
x86-64 (i7-6700 @ 3.40GHz)
taskset 4 ./residualvm monkey4-win-fr

`./configure` | `--enable-all-engines` | `--enable-all-engines --enable-optimizations` | Comment
----|------------------------------------|-----------------------------------------------------------|--------
#1271 | 24.6 ~ 25.9 | 77 ~ 81 | Same as current residualvm master
#1270 | 24.8 ~ 25.8 | 83 ~ 90 | This PR
(n/a) | 25 ~ 26 | 77 ~ 81 | This PR but requesting a 16bits mode from SDL

aquadran:
ok, good to know. when all issues with tinygl in emi will be fixed to state to quality like it was in middle of gsoc, we re-test speed check. if it will be no speed regression we will switch to 32bit mode.

vpelletier:
> when all issues with tinygl in emi will be fixed to state to quality like it was in middle of gsoc, we re-test speed check.

Are the issues you mention coming from this PR ? I guess I'll need to actually play EMI through at some point :) .

aquadran:
@vpelletier nope, I mean issues introduced while second part of gsoc task. there was no drawing issues in grim/emi before that. moslty issues with layers for blitting and transparency/alpha blending. grim engine was not fully adopted to latest changes in tinygl. emi ship scene use a lot alpha blending which is slow to render by tinygl. each fix/adding feature to proper render transparency/alpha blending in code slow a lot in that scene. this is why I'm holding 32bit mode switch until all that issues are fixed and back to proper state like it was before.

vpelletier:
Added a commit to convert save game screenshot to 16 bits, otherwise it appears garbled on load when saved from 32bits mode.

vpelletier:
Also, screenshot feature in TinyGL mode still screen-shots the menu and not the game, but this is not a regression from these patches.

aquadran:
@vpelletier can you rebase this PR?

aquadran:
@vpelletier with which version of SDL have you tested speed? 1 or 2?

vpelletier:
```
$ ldd residualvm
[...]
        libSDL-1.2.so.0 => /usr/lib/x86_64-linux-gnu/libSDL-1.2.so.0 (0x00007f5e033c2000)
[...]
```
Should I test with SDL2 too ? How do I enable it ?

aquadran:
You can enable with configure. I was curious since SDL2 use opengl as 2D backend and you will get speed boost while flip screen. So I wanted to be sure it's SDL1 in worst speed case scenario.

aquadran:
Can you put this commit: "GRIM: Enable blending around Graphics::tglBlit calls." as separated PR with comment about it "For some reason blending happens in RGB565 even without blend enabled."
Maybe we should look into tinygl code if it's proper.

aquadran:
Regarding "SDL: Do not force 16bits video mode." If we don't force 16, then we must be sure we get 32 bit screen.

vpelletier:
> Can you put this commit: "GRIM: Enable blending around Graphics::tglBlit calls." as separated PR

Done: #1290 

> If we don't force 16, then we must be sure we get 32 bit screen.

I thought it would be nicer to preserve 16bits support, even if internally everything happens in 32 bits.

aquadran:
@vpelletier maybe force create 32bit surface internally

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -87,7 +87,7 @@ void SurfaceSdlGraphicsManager::setupScreen(uint gameWidth, uint gameHeight, boo
 	if (_fullscreen)
 		sdlflags |= SDL_FULLSCREEN;
 
-	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 16, sdlflags);
+	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 0, sdlflags);
 	if (!_screen) {
```
aquadran: In SDL2 part, something like that:
-       _screenTexture = SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_RGB565, SDL_TEXTUREACCESS_STREAMING, width, height);
+       _screenTexture = SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, width, height);
-       SDL_Surface *screen = SDL_CreateRGBSurface(0, width, height, 16, 0xF800, 0x7E0, 0x1F, 0);
+       SDL_Surface *screen = SDL_CreateRGBSurface(0, width, height, 32, 0x00FF0000, 0x0000FF00, 0x000000FF, 0xFF000000);

aquadran:
@vpelletier on macOS with SDL2 I don't see speed difference between 16 and 32 bits. I'm not able to test with SDL1 on OSX currently.
With Linux under VMWare I get a lot faster with 32bits. Similiar to native with SDL2. With 16 I get slower to something I got natively between SDL1 and SDL2.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -87,7 +87,7 @@ void SurfaceSdlGraphicsManager::setupScreen(uint gameWidth, uint gameHeight, boo
 	if (_fullscreen)
 		sdlflags |= SDL_FULLSCREEN;
 
-	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 16, sdlflags);
```
aquadran: I think we should put 32 instead 0 for now, to make sure we get 32 bits.

vpelletier: Just in case there was a misunderstanding, I tried to preserve 16 bits support smoothly (ie, support for 2 and 4 bytes pixel sizes), so on a pure "does it display garbage in 16 bits" level, 16 bits is still supported.
On the 16 bits performance level, in my tests in 16 bits mode I typically saw 16 bits at the same speed as without that patch (although I did not test all rebases).

This said, I do not know how widespread 16 its is among target systems, and if rare enough among devs that code path will get broken eventually. I can imagine 3D-capable-enough systems will be also capable of 24bits colors.

aquadran: @vpelletier can you join IRC?

PR #1271, opened 2016-07-17
GRIM, EMI: Expose show_fps in engine options tab.

vpelletier:
Description borrowed from README.

### DIFF
--- a/engines/grim/detection.cpp
+++ b/engines/grim/detection.cpp
@@ -45,8 +45,9 @@ static const PlainGameDescriptor grimGames[] = {
 };
 
 #define GAMEOPTION_LOAD_DATAUSR GUIO_GAMEOPTIONS1
+#define GAMEOPTION_SHOW_FPS GUIO_GAMEOPTIONS2
 
-#define GUI_OPTIONS_GRIME GUIO1(GAMEOPTION_LOAD_DATAUSR)
+#define GUI_OPTIONS_GRIME GUIO2(GAMEOPTION_LOAD_DATAUSR, GAMEOPTION_SHOW_FPS)
 
 static const ADExtraGuiOptionsMap gameGuiOptions[] = {
 	{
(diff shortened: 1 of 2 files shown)


engines/grim/detection.cpp
```diff
@@ -56,6 +56,15 @@ static const ADExtraGuiOptionsMap gameGuiOptions[] = {
 			false
 		}
 	},
+	{
+		GAMEOPTION_LOAD_DATAUSR,
```
bgK: You'll have to define another name for your new option, and use it in the detection entries.

vpelletier: Mmh, then maybe it's not the right way to go. That option is applicable to the whole engine rather than per datafile variant. I was a bit surprised to find the existing checkbox in detection.cpp file, but did not think too much about it.

bgK: Unfortunately that's the way engine options work. They are attached to detection entries.
May I suggest using defines so that the options are not duplicated all over the place?
It is sometimes done like that in ScummVM: https://github.com/scummvm/scummvm/blob/master/engines/mohawk/detection_tables.h#L30

vpelletier:
Reworked patch set to not duplicate options. Also, I added a patch making "engine_speed=0" understood as "no FPS limit". The code supporting removal of FPS limit was already present, only configuration file reading was preventing it from being effective.

vpelletier:
FWIW, with #1266, #1278, #1270 and #1269 merged above current master and disabling FPS limit, on set ha at default coordinates and on a `configure --enable-optimisations` I more than double the FPS I get on master: from 130 to 280 on an i5-5200U @ 2.20GHz (I'm on a laptop).

vpelletier:
Latest commit fixes an assertion failure caused by a buffer overflow on extreme FPS counts. This is not a regression, but is made exceedingly easy to trigger with disabled fps limit & easy-to-enable fps counter.

bgK:
:+1:

aquadran:
@vpelletier for speed check emi, it cover full render path in tinygl

vpelletier:
In emi, the improvement is indeed much lower: in the boat scene at game start, it goes from 48 FPS (as of current master) to 54 FPS (with all merge requests mentionned above).

aquadran:
good it's not slower :)

vpelletier:
And with #1280 (in addition to previously-mentioned PRs) I get to 62FPS on the same set in emi. Grim "ha" does not change, at 280 fps.

The next single hotspot is `uint32 col = *textureBuffer;` in `putPixelTextureMappingPerspective` if I read the assembly right, so it is memory bandwidth/latency when loading texture data.

PR #1275, opened 2016-07-17
TINYGL, GRIM: Fix various errors (mostly memory leaks) found by valgrind

vpelletier:
There are still some errors, but this gets rid of a few tens of megabytes of leaks around textures.

### DIFF
--- a/graphics/tinygl/init.cpp
+++ b/graphics/tinygl/init.cpp
@@ -237,6 +237,8 @@ void glInit(void *zbuffer1, int textureSize) {
 void glClose() {
 	GLContext *c = gl_get_context();
 
+	Graphics::Internal::tglCleanupImages();
+
 	specbuf_cleanup(c);
 	for (int i = 0; i < 3; i++)
 		gl_free(c->matrix_stack[i]);
--- a/engines/grim/gfx_tinygl.cpp
+++ b/engines/grim/gfx_tinygl.cpp
@@ -63,6 +63,7 @@ GfxTinyGL::GfxTinyGL() :
 }
 
 GfxTinyGL::~GfxTinyGL() {
+	releaseMovieFrame();
 	for (unsigned int i = 0; i < _numSpecialtyTextures; i++) {
 		destroyTexture(&_specialtyTextures[i]);
 	}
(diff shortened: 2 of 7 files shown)


graphics/tinygl/init.cpp
```diff
@@ -44,7 +44,15 @@ void initSharedState(GLContext *c) {
 void endSharedState(GLContext *c) {
 	GLSharedState *s = &c->shared_state;
 
-	free_texture(c, 0);
```
bgK: I'm not sure about that. `initSharedState` allocates texture 0. Why would endSharedState free more than texture 0?
Is it to prevent memory leaks where some client code forgot to free its textures? Wouldn't that change make those cases harder to debug?

vpelletier: There is a tangential but aggravating problem here: `tglDisposeResources` takes a `GLContext` parameter. As a result, it can only be given a prototype where `zgl.h` can be imported. But `zgl.h` imports `zdirtyrect.h`, and `tglDisposeResources` is defined in `zdirtyrect.cpp` - its prototype hence cannot be defined in `zdirtyrect.h`.

There is the same problem with `zgl.h` vs. `zblit.cpp`, but "lucklily" the implementation only works with the default context - so prototype for `Graphics::Internal::tglCleanupImages` exists.
I could call `Graphics::Internal::tglCleanupImages` from `glClose` - where I think it belongs as the image list is on `GLContext`. By calling `tglDisposeResources` from `endSharedState` it will indeed only free disposed textures (and without duplicating that logic), but then will also call `Graphics::Internal::tglCleanupImages` internally... Which may be overall correct if we came from `glClose` but is inconsistent with the `GLSharedState`/`GLContext` data structure distinction.

vpelletier: And I forgot to say that I agree that close methods in general should only cause the de-allocation of disposed elements. I came to that conclusion when deciding to call `Graphics::Internal::tglCleanupImages` and forgot to go back and question my `endSharedState` change.

vpelletier: I removed that commit, I'll see if I can get somewhere changing import dependencies...

bgK:
This looks good to me without the `endSharedState` change.

vpelletier:

PR #1278, opened 2016-07-18
Tinygl: optimise scissorpixel

vpelletier:
And avoid calling it altogether when possible.

Note: contains #1266 as it would otherwise conflict, I don't know how to do otherwise to still propose these patches.

### DIFF
--- a/graphics/tinygl/zbuffer.h
+++ b/graphics/tinygl/zbuffer.h
@@ -214,9 +214,7 @@ struct FrameBuffer {
 		writePixel(pixel, 255, rSrc, gSrc, bSrc);
 	}
 
-	FORCEINLINE bool scissorPixel(int pixel) {
-		int x = pixel % xsize;
-		int y = pixel / xsize;
+	FORCEINLINE bool scissorPixel(int x, int y) {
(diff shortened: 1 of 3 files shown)


bgK:
:+1: (for the changes new in this patchset)

vpelletier:

aquadran:
hmm, it contain changes from other PR

vpelletier:
True. Also, i've been working on zdirtyrect lately (not sure anything conclusive will come out of it yet) and I changed my mind on how scissorPixel should be optimised: zdirtyrect makes extensive use of it, and I only recently realised it is actually mostly dead code so far (it uses whole-screen blits unless enabled specifically, and it's not enabled).
And when enabling, performance becomes abysmal just because of the modulo in scissorPixel. So I made all callers know (and provide) the 2D coordinates to scissorPixel, so it can directly compare without any operation - and as a result, I think it is overkill to control just 4 comparisons with a with a template-level flag.

vpelletier:
I rebased this pull-request over master, and pushed the optimisation of `scissorPixel` further, to prepare for enabling dirty rectangle mechanism by default (...which may still require a lot of work, but I'm not sure where to discuss it).

PR #1279, opened 2016-07-18
TINYGL: reduce code duplication in ztriangles (was: [RFC] TINYGL: unroll some loops)

vpelletier:
Some loops are written in a way which suggests loop-unrolling was intended, but code was re-rolled. This change unrolls these loops.

Note: I did not see any obvious performance difference. If anything, this makes reading the code easier by reducing repetitions, and maybe fixing tiny (runs of 3 pixels) rendering approximations.

This brings #1270 changes because of conflicts.

### DIFF
--- a/graphics/tinygl/ztriangle.cpp
+++ b/graphics/tinygl/ztriangle.cpp
@@ -75,6 +75,14 @@ FORCEINLINE static void putPixelDepth(FrameBuffer *buffer, int buf, unsigned int
 	z += dzdx;
 }
 
+template <bool kDepthWrite, bool kAlphaTestEnabled, bool kEnableScissor, bool kBlendingEnabled>
+FORCEINLINE static void putPixelShadow(FrameBuffer *buffer, int buf, unsigned int *pz, int _a, int x, int y, unsigned int &z, int color, int &dzdx, unsigned char *pm) {
+	if ((!kEnableScissor || !buffer->scissorPixel(x + _a, y)) && buffer->compareDepth(z, pz[_a]) && pm[_a]) {
+		buffer->writePixel<kAlphaTestEnabled, kBlendingEnabled, kDepthWrite>(buf + _a, color, z);
+	}
+	z += dzdx;
+}
+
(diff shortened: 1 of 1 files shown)


vpelletier:
...and this actually decreases performance. Funny.

vpelletier:
With this last patch, an initial 62FPS in first EMI scene (without this pull request) is increased to 65 FPS (with unrolling disabled) instead of decreasing to 58FPS.

Should I track more precisely which unrolls are beneficial and detrimental ?
Once the unrolls to drop are known, should I keep them around or just delete them ?

vpelletier:
The commit disabling manual loop unrolling breaks MYST 3 (in a rather funny, wobbly way).

vpelletier:
And the regression is what caused the performance improvement: in the `(kInterpST || kInterpSTZ)` branch, s, t, dsdx, dtdx and fz were not updated at all, causing the texture rendering glitches. Re-enabling their computation, the speed is at 62FPS without any unrolling (so overall, manual unrolling still costs 4 FPS).
I fix this in latest version of this PR.
Maybe I should split the second commit further ?

vpelletier:
FWIW, I am scavenging commits which do not add 32bits support per-se from this pull request (and other changes I initially prepared after that branch), so that they can be merged independently and I can keep this branch floating over master until the surrounding situation makes a merge acceptable.

vpelletier:
Now this PR only contains a single commit, which does not depent on 32bits work but mostly reduces code duplication (which in turn makes 32bits work a much simpler patch).

graphics/tinygl/ztriangle.cpp
```diff
@@ -403,6 +403,8 @@ void FrameBuffer::fillTriangle(ZBufferPoint *p0, ZBufferPoint *p1, ZBufferPoint
 			{
 				if (kDrawLogic == DRAW_DEPTH_ONLY ||
 						(kDrawLogic == DRAW_FLAT && !(kInterpST || kInterpSTZ))) {
+#define UNROLL_putPixelDepth(_a) putPixelDepth<kDepthWrite, kEnableScissor>(this, buf, pz, _a, x, y, z, dzdx)
```
aquadran: No macros please :)

vpelletier:
Now without macros.

PR #1280, opened 2016-07-19
GRAPHICS: Optimise PixelBuffer::getValueAt and PixelBuffer::setPixelAt.

vpelletier:
Avoid doing a lot of operations just to move 2 or 4 bytes around.

### DIFF
--- a/graphics/pixelbuffer.cpp
+++ b/graphics/pixelbuffer.cpp
@@ -86,20 +86,6 @@ void PixelBuffer::clear(int length) {
 	memset(_buffer, 0, length);
 }
 
-void PixelBuffer::setPixelAt(int pixel, uint32 value) {
-#if defined(SCUMM_BIG_ENDIAN)
-	byte *buffer = _buffer + pixel * _format.bytesPerPixel;
-	for (int i = 0; i < _format.bytesPerPixel; ++i) {
-		buffer[i] = value >> ((_format.bytesPerPixel - i - 1) * 8) & 0xFF;
-	}
-#elif defined(SCUMM_LITTLE_ENDIAN)
-	byte *buffer = _buffer + pixel * _format.bytesPerPixel;
-	for (int i = 0; i < _format.bytesPerPixel; ++i) {
-		buffer[i] = value >> (i * 8) & 0xFF;
-	}
-#endif
-}
-
 void PixelBuffer::copyBuffer(int thisFrom, int otherFrom, int length, const PixelBuffer &buf) {
 	if (buf._format == _format) {
 		memcpy(_buffer + thisFrom * _format.bytesPerPixel, buf._buffer + otherFrom * _format.bytesPerPixel, length * _format.bytesPerPixel);
@@ -110,24 +96,6 @@ void PixelBuffer::copyBuffer(int thisFrom, int otherFrom, int length, const Pixe
 			setPixelAt(i + thisFrom, r, g, b);
 		}
 	}
-}
-
-uint32 PixelBuffer::getValueAt(int i) const {
-#if defined(SCUMM_BIG_ENDIAN)
-	byte *buffer = _buffer + i * _format.bytesPerPixel;
-	uint32 p = buffer[0] << ((_format.bytesPerPixel - 1) * 8);
-	for (int l = 1; l < _format.bytesPerPixel; ++l) {
-		p = p | (buffer[l] << (8 * (_format.bytesPerPixel - l - 1)));
-	}
-	return p;
-#elif defined(SCUMM_LITTLE_ENDIAN)
-	byte *buffer = _buffer + i * _format.bytesPerPixel;
-	uint32 p = buffer[0];
-	for (int l = 1; l < _format.bytesPerPixel; ++l) {
-		p = p | (buffer[l] << (8 * l));
-	}
-	return p;
-#endif
 }
 
 PixelBuffer &PixelBuffer::operator=(const PixelBuffer &buf) {
--- a/graphics/pixelbuffer.h
+++ b/graphics/pixelbuffer.h
@@ -24,6 +24,8 @@
 #define GRAPHICS_PIXELBUFFER_H
 
 #include "common/types.h"
+#include "common/endian.h"
+#include "common/textconsole.h"
 
 #include "graphics/colormasks.h"
 #include "graphics/pixelformat.h"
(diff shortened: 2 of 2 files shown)


aquadran:
I'm curious what profile tool says about it? I never liked that code since it introduced slow down.

vpelletier:
Before this change, `setPixelAt` was in 3rd position at 4.5% IIRC, and getValueAt was in 5th.
After this change, it is visible in https://github.com/residualvm/residualvm/pull/1271#issuecomment-233504153 : 2.8% and 1.3% respectively. Note that as these percentages are over a different number of frames per second (ie, these methods ended up being called more than before), they cannot be simply compared with the original percentages.

I wanted to inline them both, but it seems the method then has to be defined in the header (at least, before other other methods reference it), and I don't like so much implementing method in header files. So I tried moving other methods in the .cpp file, but just got more errors and got bored.

vpelletier:
Oh wait, my mistake: I made that screen dump before this change. The ranks I remembered were with grim, likely.
Here it is after this change, along with hotest methods overall and other PixelBuffer methods:
```
Samples: 940K of event 'cycles:ppp', Event count (approx.): 6922314263, DSO: residualvm [z]
Overhead  Symbol
  21,21%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, false, false, false, true, false>
  14,22%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, false, false, false, false, false>
   3,94%  [.] Graphics::PixelBuffer::copyBuffer
   3,41%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, true, false, false, false, false>
   2,39%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, false, 2, true, false, false, false, false>
   2,32%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, true, true, false, true, false>
   2,07%  [.] TinyGL::FrameBuffer::clearRegion
   1,73%  [.] TinyGL::FrameBuffer::genOffscreenBuffer
   1,32%  [.] Graphics::Internal::tglBlit<true, true, true, false>
[...]
   0,05%  [.] Graphics::PixelBuffer::getValueAt
[...]
   0,03%  [.] Graphics::PixelBuffer::create
[...]
   0,02%  [.] Graphics::PixelBuffer::operator=
```
`setPixelAt` is nowhere to be found - maybe gcc was smart enough to inline it ?

vpelletier:
Weirder and weirder: after posting my previous comment and switching back to `perf top` the lines changed a lot although I enabled "zero samples on refresh". I'm getting lost.

graphics/pixelbuffer.cpp
```diff
@@ -88,14 +88,34 @@ void PixelBuffer::clear(int length) {
...
 #if defined(SCUMM_BIG_ENDIAN)
-	byte *buffer = _buffer + pixel * _format.bytesPerPixel;
-	for (int i = 0; i < _format.bytesPerPixel; ++i) {
-		buffer[i] = value >> ((_format.bytesPerPixel - i - 1) * 8) & 0xFF;
+	switch (_format.bytesPerPixel) {
+	case 2:
+		((uint16 *) _buffer)[pixel] = (uint16) ((value << 8) | (value >> 8));
+		break;
+	case 4:
+		((uint32 *) _buffer)[pixel] = (uint32) (((value & 0xFF00FF00) >> 8) | ((value & 0x00FF00FF) << 8));
```
bgK: Is this line supposed to be a 32 bits endian swap?
If so, I would suggest using the SWAP_BYTES_32 macro, which will use a compiler builtin in most cases.

bgK:
If the cost of calling and returning from these functions is important, it may be interesting to try making the compiler able to inline them by moving them to a header. Possibly removing the default implementation to reduce the generated code size, so that they are more likely to be inlined.

In an unrelated note to this pull request, I'm surprised `setPixelAt` and `getValueAt` need to do endian conversion. I would expect the `value` argument to be in the pixel format defined by `_format`.

graphics/pixelbuffer.cpp
vpelletier: Yes, it was supposed to be a byte swap, but I missed swapping the 16 bit around. Thanks for the macro pointer.

vpelletier:
With latest version, `setPixelAt` and even more `getValueAt` went a lot lower in percentage. It is not enough to gain a clear extra frame per second with emi though.
```
   0,96%  residualvm  [.] Graphics::PixelBuffer::setPixelAt
   0,71%  residualvm  [.] Graphics::PixelBuffer::copyBuffer
   0,02%  residualvm  [.] Graphics::PixelBuffer::operator=
   0,02%  residualvm  [.] Graphics::PixelBuffer::PixelBuffer
   0,01%  residualvm  [.] Graphics::PixelBuffer::~PixelBuffer
   0,01%  residualvm  [.] Graphics::PixelBuffer::create
   0,01%  residualvm  [.] Graphics::PixelBuffer::PixelBuffer
   0,01%  residualvm  [.] Graphics::PixelBuffer::PixelBuffer
   0,00%  residualvm  [.] Graphics::PixelBuffer::free
   0,00%  residualvm  [.] Graphics::PixelBuffer::getValueAt
```
I also discovered 3-bytes support is needed for emi. I sadly have no big-endian machine at hand which would run residual to check code endianness.

bgK:
:+1:

vpelletier:
Continuing my hijack of my own PR thread to track performance from all my other PRs...

Now emi is at 65~67FPS with latest change sets, with the following output from `perf top`:
```
  30,69%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, false, false, false, true, false>
  17,31%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, false, false, false, false, false>
   4,30%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, true, false, false, false, false>
   3,23%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, false, 2, true, false, false, false, false>
   3,01%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, true, true, false, true, false>
   1,36%  [.] Graphics::Internal::tglBlit<true, true, true, false>
   1,32%  [.] Math::operator*<4, 1, 4>
   1,28%  [.] TinyGL::glopVertex
   1,08%  [.] TinyGL::FrameBuffer::blitOffscreenBuffer
   0,80%  [.] Graphics::PixelBuffer::copyBuffer
   0,74%  [.] TinyGL::gl_draw_triangle
   0,62%  [.] TinyGL::gl_add_op
   0,56%  [.] Grim::EMIModel::prepareForRender
   0,54%  [.] Grim::GfxTinyGL::drawEMIModelFace
   0,53%  [.] Graphics::RasterizationDrawCall::computeDirtyRegion
   0,46%  [.] Grim::EMIModel::updateLighting
   0,42%  [.] TinyGL::FrameBuffer::fillTriangle<false, true, false, false, 4, false, false, false, false, false>
   0,31%  [.] TinyGL::gl_transform_to_viewport
   0,27%  [.] TinyGL::tglPresentBufferSimple
   0,27%  [.] Math::Matrix<4, 4>::Matrix
   0,26%  [.] Graphics::PixelBuff
```

And grim is around 300~330fps:
```
  21,28%  [.] TinyGL::FrameBuffer::fillTriangle<true, true, false, true, 2, true, true, false, false, false>
   6,38%  [.] TinyGL::FrameBuffer::blitOffscreenBuffer
   5,20%  [.] TinyGL::gl_shade_vertex
   2,73%  [.] TinyGL::tglPresentBufferSimple
   2,03%  [.] TinyGL::glopVertex
   1,66%  [.] TinyGL::gl_add_op
   1,46%  [.] TinyGL::FrameBuffer::fillTriangle<false, true, false, false, 4, false, false, false, false, false>
   1,17%  [.] TinyGL::gl_draw_triangle
   0,92%  [.] Graphics::RasterizationDrawCall::applyState
   0,74%  [.] Graphics::RasterizationDrawCall::computeDirtyRegion
   0,64%  [.] Graphics::RasterizationDrawCall::captureState
   0,64%  [.] Graphics::PixelFormat::colorToARGB
   0,61%  [.] Graphics::RasterizationDrawCall::execute
   0,56%  [.] TinyGL::gl_transform_to_viewport
   0,42%  [.] Graphics::RasterizationDrawCall::RasterizationDrawCall
   0,41%  [.] Grim::GfxTinyGL::drawModelFace
   0,38%  [.] Math::Matrix<4, 4>::Matrix
```

PR #1281, opened 2016-07-19
TINYGL: Actually set memory to the requested legnth.

vpelletier:
4 is 1 << 2, not 1 << 3.
Fixes display corruption (especially visible in Myst 3) where the right
half of the screen contains data from previous frame.

### DIFF
--- a/graphics/tinygl/zbuffer.cpp
+++ b/graphics/tinygl/zbuffer.cpp
@@ -47,7 +47,7 @@ void memset_s(void *adr, int val, int count) {
 	p = (unsigned int *)adr;
 	v = val | (val << 16);
 
-	n = count >> 3;
+	n = count >> 2;
 	for (int i = 0; i < n; i++) {
 		p[0] = v;
 		p[1] = v;


bgK:
Good catch !

That function still looks suspicious to me due to `int` being 64 bit on some platforms.

vpelletier:
Actually, the original code is correct if count is in 16bits words, and may change breaks consistency between the right-shift and the mask further below. I think function is overcomplex: it should just call memset_l and just handle the case of an odd count.

Also, Framebuffer::clear calling memset_s makes the assumption that the buffer is in a 16-bits color mode - which is not true for me anymore. So I think there are 2 things here:
- code simplification (and correction for 64bits ints) in memset_* which should belong here
- framebuffer bytes-per-pixel assumption removal, to be made part of #1270 .

vpelletier:
After working on the latter fix (just pushed), I made thecode avoid calling memset_{s,l} in practice. So I do not intend to work further on these functions (at least not in near future).

PR #1286, opened 2016-07-21
TINYGL: Unroll FrameBuffer::blitOffscreenBuffer.

vpelletier:
This method is used to blit large buffers, so a large unrolling is still
beneficial. As it is the second hottest-spot in grim, the frame rate
increase is most significant there (compared to emi).

I left the `// TODO: could be faster, probably.` comment as this method is still rather costly.

### DIFF
--- a/graphics/tinygl/zbuffer.cpp
+++ b/graphics/tinygl/zbuffer.cpp
@@ -210,20 +210,67 @@ void FrameBuffer::clearRegion(int x, int y, int w, int h, int clearZ, int z, int
 	}
 }
 
+inline static void blitPixel(uint8 offset, unsigned int *from_z, unsigned int *to_z, unsigned int z_length, byte *from_color, byte *to_color, unsigned int color_length) {
+	const unsigned int d = from_z[offset];
+	if (d > to_z[offset]) {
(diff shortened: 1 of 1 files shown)


graphics/tinygl/zbuffer.cpp
```diff
@@ -170,17 +170,65 @@ void FrameBuffer::clearRegion(int x, int y, int w, int h, int clearZ, int z, int
 
 void FrameBuffer::blitOffscreenBuffer(Buffer *buf) {
 	// TODO: could be faster, probably.
+#define UNROLL_blit(_a) {\
```
aquadran: No macros, please :). We moved away in tinygl. They are hard to debug. Maybe we could use templates.

vpelletier:
I pushed a separate commit to get rid of the macro. Is it the code which is intended ? Although I can guess having multi-statement macros can cause weird problems, I don't like so much the parameter repetition, and the many lines of almost-identical conflicts which can come with it (and which do come with it in #1279 when I need to rebase my 32bits work).

aquadran:
ok, I think it's fine

vpelletier:
Pushed squashed. I've mostly finished reworking the other patch set too.

PR #1287, opened 2016-07-23
TINYGL:  Enable dirtyrect by default.

vpelletier:
Once #1295 and #1335 are merged, I believe dirty rectangles can finally be enabled by default ! (this PR contains both, for testing ease)

Actually, I think about getting rid of the non-dirtyrect code paths in TinyGL. Not sure it would be a good idea to include in this PR, though.

I gave up on my attempts to rework rectangle merging algorithm: current code is simple where my code is complex and brittle - and both perform just the same.

There is one part where current code would likely benefit from improvements: using a diff algorithm against draw call lists, instead of current `a++ == b++`-style comparison.

Also, a bit of trivia: on my T450s, TinyGL now give more FPS than OpenGL (i5-4300U 1.9GHz + HD4400), both with same residualvm built with `--enable-optimisations`. Both being in the few hundreds, it does not mean much but... I found it funny and kind of rewarding.

### DIFF
--- a/graphics/tinygl/init.cpp
+++ b/graphics/tinygl/init.cpp
@@ -237,7 +237,7 @@ void glInit(void *zbuffer1, int textureSize) {
 	c->_currentAllocatorIndex = 0;
 	c->_drawCallAllocator[0].initialize(kDrawCallMemory);
 	c->_drawCallAllocator[1].initialize(kDrawCallMemory);
-	c->_enableDirtyRectangles = false;
+	c->_enableDirtyRectangles = true;
 
 	Graphics::Internal::tglBlitResetScissorRect();
 }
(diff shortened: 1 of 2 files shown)


bgK:
I did a quick test of Myst III with this branch. It works very well :+1:

vpelletier:
@bgK : When you say "very well", do you mean you saw improvements ? I believe Myst III basically always redraws the whole screen, so I would not expect much performance improvement there.

And if you mean you did not notice any regression, that's of course very valuable to know :) .

I would like to get more feedback on these 3 pull requests (or even everything on this one, as it has everything) before merging. Maybe this week-end ?

bgK:
>When you say "very well", do you mean you saw improvements ? I believe Myst III basically always redraws the whole screen, so I would not expect much performance improvement there.

I did not do serious performance tests. It felt faster than usual, but it could as well have been caused by some of your earlier changes, a newer compiler, or the fact I was using my laptop instead of my desktop.

PR #1290, opened 2016-07-23
GRIM: Enable blending around Graphics::tglBlit calls

vpelletier:
For some reason blending happens in RGB565 even without blend enabled.

### DIFF
--- a/engines/grim/gfx_tinygl.cpp
+++ b/engines/grim/gfx_tinygl.cpp
@@ -1044,7 +1044,14 @@ void GfxTinyGL::drawBitmap(const Bitmap *bitmap, int x, int y, uint32 layer) {
 	Graphics::BlitImage **b = (Graphics::BlitImage **)bitmap->getTexIds();
 
 	if (bitmap->getFormat() == 1) {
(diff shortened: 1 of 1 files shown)


engines/grim/gfx_tinygl.cpp
```diff
@@ -1044,7 +1044,14 @@ void GfxTinyGL::drawBitmap(const Bitmap *bitmap, int x, int y, uint32 layer) {
 	Graphics::BlitImage **b = (Graphics::BlitImage **)bitmap->getTexIds();
 
 	if (bitmap->getFormat() == 1) {
+		if (bitmap->getHasTransparency()) {
```
aquadran: Can you add comment here?

vpelletier: You mean on the `if` itself ? I copied it from `gfx_opengl.cpp` (without completely sync'ing the method either...).

aquadran: I mean add comment close that lines.

vpelletier: Sorry for comming back late, I just woke up.

PR #1294, opened 2016-07-28
Tinygl memory

vpelletier:
More issues found by valgrind while working on dirty rectangles.
Also, contains a better implementation of texture/image cleanup (previous attempt was rejected in #1275 ).

### DIFF
--- a/graphics/tinygl/zdirtyrect.h
+++ b/graphics/tinygl/zdirtyrect.h
@@ -156,7 +156,7 @@ public:
 	};
 
 	BlittingDrawCall(BlitImage *image, const BlitTransform &transform, BlittingMode blittingMode);
-	virtual ~BlittingDrawCall() { }
+	virtual ~BlittingDrawCall();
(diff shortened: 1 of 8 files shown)


aquadran:
👍

graphics/tinygl/zdirtyrect.h
```diff
@@ -156,7 +156,7 @@ class BlittingDrawCall : public DrawCall {
 	};
 
 	BlittingDrawCall(BlitImage *image, const BlitTransform &transform, BlittingMode blittingMode);
-	virtual ~BlittingDrawCall() { }
+	~BlittingDrawCall();
```
bgK: The destructor should still be virtual here since the base class has a vtable.

vpelletier: Thanks for keeping an eye out for C++ mistakes. For some reason I remembered "virtual" methods as being left unimplemented, but RTFM'ing a bit I see that's not what this keyword is about. Fixed.

PR #1298, opened 2016-08-22
MYST3: Load scripts from a ResidualVM provided file

bgK:
This changeset's goal is to ensure the most up to date game scripts are used in Myst III.
When the DVD version of Myst III was released the game scripts contained in the game executable were updated fixing a few bugs. However, Presto only released a patch to update the French and German Mac CD versions to those new scripts. The other CD versions are left with known bugs.

This PR adds a dev tool called 'create_myst3' that extracts the game scripts from various game executables and stores them in a file named 'myst3.dat'. This file is distributed with ResidualVM. ResidualVM uses it to load the game scripts instead of loading them from the user provided game executable, as previously.

As a side effect script loading is much simplified, with the removal of PEF and UPX decompression, and SafeDisc decryption.

This will allow in the future not to require the game executable. Getting the game executable is a problem for some users because it is recognized as a false positive malware by some antivirus software.
For now, the game executable is still required for detection, but not otherwise accessed.

This approach is commonly used in ScummVM, but I want to make sure it is ok with the ResidualVM team.


bgK:
I'll merge this in a few days unless somebody has comments.

aquadran:
@bgK I want discuss about it first

aquadran:
:+1:

PR #1301, opened 2016-09-15
GRIM: Enable the Global Game Menu

bgK:
In the grim engine, ResidualVM's Global Game Menu is overridden to show the game menu instead. Now that ResidualVM supports several games, I think that the GMM override is no longer justified.

This re-enables the GMM, allowing to return to the launcher and start another game.


aquadran:
👍

Issue #1300, opened 2016-09-04
EMI-Demo: Game crashes at start with "Unknown pixelformat"


immi-h:
[patch.zip](https://github.com/residualvm/residualvm/files/453844/patch.zip)
When starting EMI-Demo, the game almost immediately goes into the debugging console, saying
"ERROR: Unknown pixelformat: Bpp: 4 RBits: 8 GBits: 8 BBits: 8 ABits: 0 RShift: 16 GShift: 8 BShift: 0 AShift: 0!"

This only happens when under options, the renderer is OpenGL. (No problems with "OpenGL with shaders" or with "software")

With the attached patch, the problem disappears, but I guess that's not what should be done. (Without the patch, bpp is set to 32 on my system.)

If you have difficulties reproducing the error, I can try to find the real problem, but I'd need some help. (For the moment, I don't know where to look.)

vpelletier:
Hello. Would you be on a big-endian macine ?
The error message comes from [grim/gfx_opengl.cpp](https://github.com/residualvm/residualvm/blob/d467980499120a7cd3796e8b5d2e8715f273317d/engines/grim/gfx_opengl.cpp#L1587). The color description in your error shows it's 24bits (32bits with no alpha), but the component shifts are oppposite to the one tested in the first branch of the `if` - so it looks like just lack of big-endian support.

Could you try adding a branch in that `if` like this:
```c++
        } else if (frame->format == Graphics::PixelFormat(4, 8, 8, 8, 0, 16, 8, 0, 0)) {
                format = GL_RGBA;
                dataType = GL_UNSIGNED_INT_8_8_8_8;
```
(note `format` is different from the one in existing 24bits-color branch)

immi-h:
This added `if` indeed solves the problem. Thanks.
I was also puzzled by the component order in the error message. However, endianness doesn't seem to be the reason. I just double checked: My computer really is little endian.

vpelletier:
> However, endianness doesn't seem to be the reason. 

Yes, it looks like I assumed to much: this function receives data from game files (a video).

It's not clear to me why this error message wouldn't be triggered just by requesting a 16bits mode though.

Also, does the video (Aspyr logo video, as per a comment in the source code) have the same color with both patches (your and mine) ? (hopefully it has separate red-ish and blue-ish areas so you can tell)

immi-h:
Oops, you're right. Your patch makes the video have wrong colors.

[...]

PR #1302, opened 2016-09-17
GRIM: Fix support for reverse-order 32bits video frames in OpenGL

vpelletier:
Make shader renderer fail the same way as opengl renderer on unexpected pixel format, which also allows getting rid of the "swizzle" uniform by letting opengl handle it ( @Botje : please confirm there was no other intent, like maybe some android workaround).

Add support for RGBA REV pixel format, so opengl and shader renderers do not fail as reported in #1300 .

Also, request quicktime high-color videos to use the 8/16/24 pixel format (untested as I could not identify a quicktime video in the EMI EN demo linked from residualvm website).

Patch set tested against:
- shaders + grim-win-fr (intro video is RGB, so it's a different code path anyway)
- shaders + monkey4-win-fr (sinking ship & arrival on melee videos are RGBA REV, hitting the new code path)
- opengl + monkey4-win-fr (likewise)
- shaders + monkey4-demo-en (intro video is  RGBA REV, hitting the new code path)
- opengl + monkey4-demo-en (likewise)

### DIFF
--- a/engines/grim/gfx_opengl.cpp
+++ b/engines/grim/gfx_opengl.cpp
@@ -1587,6 +1587,9 @@ void GfxOpenGL::prepareMovieFrame(Graphics::Surface *frame) {
 	if (frame->format == Graphics::PixelFormat(4, 8, 8, 8, 0, 8, 16, 24, 0)) {
 		format = GL_BGRA;
 		dataType = GL_UNSIGNED_INT_8_8_8_8;
+	} else if (frame->format == Graphics::PixelFormat(4, 8, 8, 8, 0, 16, 8, 0, 0)) {
+		format = GL_BGRA;
+		dataType = GL_UNSIGNED_INT_8_8_8_8_REV;
 	} else if (frame->format == Graphics::PixelFormat(2, 5, 6, 5, 0, 11, 5, 0, 0)) {
 		format = GL_RGB;
 		dataType = GL_UNSIGNED_SHORT_5_6_5;
--- a/engines/grim/gfx_opengl_shaders.cpp
+++ b/engines/grim/gfx_opengl_shaders.cpp
@@ -48,7 +48,7 @@
 #include "common/system.h"
 #include "common/textconsole.h"
 
-#ifdef USE_OPENGL_SHADERS
+#if defined(USE_GLES2) || defined(USE_OPENGL_SHADERS)
 
 #include "graphics/surface.h"
 #include "graphics/pixelbuffer.h"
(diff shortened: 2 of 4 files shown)


vpelletier:
The jenkins failure looks legitimate:
```
    C++      engines/grim/gfx_opengl_shaders.o
engines/grim/gfx_opengl_shaders.cpp: In member function 'virtual void Grim::GfxOpenGLS::prepareMovieFrame(Graphics::Surface*)':
engines/grim/gfx_opengl_shaders.cpp:1903:15: error: 'GL_UNSIGNED_INT_8_8_8_8' was not declared in this scope
   frameType = GL_UNSIGNED_INT_8_8_8_8;
               ^
engines/grim/gfx_opengl_shaders.cpp:1906:15: error: 'GL_UNSIGNED_INT_8_8_8_8_REV' was not declared in this scope
   frameType = GL_UNSIGNED_INT_8_8_8_8_REV;
               ^
make: *** [Makefile.common:106: engines/grim/gfx_opengl_shaders.o] Error 1
```
But I don't see yet where it comes from. It does build on my machine, with also:
```
Checking for GL Extension Wrangler (GLEW)... yes
[...]
Checking for OpenGL... yes (shaders enabled)
[...]
Backend... sdl (1.2.15), savegame timestamp
```
in `configure` output.

bgK:
The CI job does two builds. One regular Linux build, and an Android build. It's the second one that fails because OpenGL ES 2 does not have some of the pixel formats used here.

vpelletier:
Updated patch stack builds on both x86_64 linux (debian sid) and android (android-ndk-r12b[1]). Both swizzle is required by and build failure is caused by GLES2 support. I split original swizzle operation to handle endianness and component swapping separately, as needed for `_REV` support.

Re-tested against:

- shaders + grim-win-fr (intro video is RGB, so it's a different code path anyway) on android & linux
- shaders + monkey4-win-fr (sinking ship & arrival on melee videos are RGBA REV, hitting the new code path) on linux (could not complete the first puzzle on android, impossible to use the brasero - "use" button triggers only "look at")
- opengl + monkey4-win-fr (likewise) on linux
- shaders + monkey4-demo-en (intro video is RGBA REV, hitting the new code path) on android & linux
- opengl + monkey4-demo-en (likewise) on linux

[1] Requires a `4.9.x` -> `4.9` symlink in `sources/cxx-stl/gnu-libstdc++` so `libsupc++.a` can be found.

bgK:
The changes look ok to me, fine for merging.

I'm just a bit curious. The video decoders allow to choose the output pixel format. So why not telling them to output in a format that's directly supported?

Say:
Little Endian: Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24) => GL_RGBA / GL_UNSIGNED_BYTE
Big Endian: Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0) => GL_RGBA / GL_UNSIGNED_BYTE

vpelletier:
I don't know how this works enough. I did a few tries when I noticed the different colors on android, but maybe this was before I checked GLES2 doc and changed back to GL_RGBA. So I thought it would be better to just support more formats.

Probably on a different topic, I see the background animations (ship, brasero) on the first monkey4 non-demo puzzle have the wrong color. It's not a regression from this change. I'm not sure how these are rendered.

vpelletier:
I successfully completes the first uzzle in "shaders + monkey4-win-fr" on android (took me some time to discover all the input method features), and the video after it has correct colors.

vpelletier:
And I was wrong in the format used in emi-win-fr vs. emi-demo-win: the intro (starting with "Somewhere off the coast of Mêlée Island...") is RGB888 in the demo ad BGR888 in FR.

vpelletier:
Rebased, without the commit which was hinting the video decoder to use a specific color format: I do not know how to choose one format over others, nor how many other video decoder may need hinting. This has been bugging me and preventing me from merging. So I prefer to just drop this tiny, non-essential part from this PR and merge it - as it actually fixes at least one reported bug.

I intend to merge when checks pass.

PR #1305, opened 2016-10-12
STARK: Add support for screen shake

firesock:


bgK:
Thanks for your contribution. In the case of the Stark engine I only accept new features when the behavior is accurate with regards to the original.

In this case, scene rumbling should work as follows:
- The first argument for opRumbleScene is indeed the duration in ms.
- The second argument is "suspend". It indicates whether the script should be paused for the duration of the rumbling or if it should continue running. (If true, Script->pause should be called)

During the rumble effect, each layer should move independently. The original implements that in `Location::listRenderEntries` which is called on each frame. For each individual layer when rumbling is active, the game generates two random values between -1 and 1. These values are used to temporarily offset the x and y coordinates of the scroll property of the layers. That way the render entries' positions are computed with a slight offset. The scroll property is then restored to its original value.

If you wish to contribute to the Stark engine, I can give you access to my disassembly of the game, and you are welcome to ask me questions regarding how things are done in the original engine.

firesock:
I see, I was mostly working with watching and guessing for this implementation. I'll raise another PR when I get this working that way then.

It would be useful to have the disassembly, if you don't mind?

bgK:
Ok, I have contacted you by email. I'm closing this PR while you work on the new implementation.

PR #1313, opened 2016-11-06
BUILD: Use SDL2 by default

bgK:
Our SDL 2 support has been working well for me for some time now. Now that ScummVM switched to using it by default over SDL 1.2, I think it's time we do the same.

Do you know of bugs that are specific to the SDL 2 version of ResidualVM?


aquadran:
let me test on my setups

bgK:
With the recent ScummVM sync, this is now the default.

PR #1314, opened 2016-11-06
COMMON: Synch changes to {game,advancedDetector}.{h,cpp} from ScummVM

somaen:
While attempting to copy over the Wintermute engine code, I noticed that ScummVM renamed some variables in the detection-system.

So, I copied over the relevant files from ScummVM-master, added all the changes, and resolved the compile-errors I found in the other engines.

AdvancedDetector had a single ResidualVM-specific string (The string saying where to file missing detection entries says "ResidualVM team" in ResidualVM, and "ScummVM Team" in ScummVM), which I did NOT add to this patch, so you will find that diff between ScummVM and this, but otherwise the files are exact copies.

### DIFF
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -41,8 +41,8 @@ static GameDescriptor toGameDescriptor(const ADGameDescription &g, const PlainGa
 		title = g.extra;
 		extra = "";
 	} else {
-		while (sg->gameid) {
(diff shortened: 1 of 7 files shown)


bgK:
:+1:

aquadran:
I'm a middle of sync, which I stuck at some point, but in this case I'll resume it.
When you need sync with ScummVM let me know.

PR #1315, opened 2016-11-15
STARK: Add support for opRumble

firesock:
Implementation of opRumble based on the original version. I have 2 questions here:

- @bgK mentioned in #1305 that the random values are between -1 and 1. Am I missing something here, I can only see offset in each axis of either -1 or 0 and restoring in the end?

- Rumbling is still quite fast in a visual comparison. This seems down to the value of millisecondsPerGameLoop, which I think is set in w_module.dll(0x10029372) to a static value of 33? Is it worth either increasing our mainLoop delayMillis call or something more complicated in updateDisplayScene?

### DIFF
--- a/engines/stark/resources/command.cpp
+++ b/engines/stark/resources/command.cpp
@@ -101,7 +101,7 @@ Command *Command::execute(uint32 callMode, Script *script) {
 	case kGoto2DLocation:
 		return opGoto2DLocation(_arguments[0].stringValue, _arguments[1].stringValue);
 	case kRumbleScene:
-		return opRumbleScene(_arguments[1].intValue, _arguments[2].intValue);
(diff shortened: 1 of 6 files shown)


bgK:
Looks good, thanks. Please initialize _rumbleFramesRemaining in the Location class constructor, then it will be fine to merge.

* You are right about the random values, they are between -1 and 0 only.
* For now, the game runs too fast. We need to add a frame limiter to have a stable framerate (probably something like what is used in Myst III, see https://github.com/residualvm/residualvm/commit/6cb71cd52a9fe0b6a8822d0c910296fcf86bbf79). The value of `millisecondsPerGameLoop` is initialized to 33 in the original engine but is then updated according to the frame duration.

firesock:
Ah I see, thanks. I'll leave the timing alone for now then.

PR #1316, opened 2016-11-20
Allow to control V-Sync

bgK:
This adds a option to enable or disable waiting for V-Sync to refresh the screen. It is enabled by default.

This fixes the very noticeable tearing in Myst III.


raziel-:
Just for information...is the tearing fixed by using V-Sync by default or by turning it off?

bgK:
The tearing is fixed by waiting for V-Sync to update the screen so that the frame is fully consistent. So by having the option enabled.
The option can be disabled for situations where the frames are not rendered fast enough, and are late for V-Sync. In that case, when V-Sync is enabled the missed frames are very noticeable.

common/system.h
```diff
@@ -323,6 +323,13 @@ class OSystem : Common::NonCopyable {
 		kFeatureSideTextures,
 
 		/**
+		 * If supported, this feature flag can be used to enable or disable
+		 * waiting for vertical sync before refreshing the screen to reduce
+		 * tearing.
+		 */
+		kFeatureVSync,
+
```
aquadran: please mention it's residualvm specific

PR #1326, opened 2017-03-26
STARK: Implement opFadeScene

firesock:


engines/stark/resources/location.cpp
```diff
@@ -65,6 +67,20 @@ void Location::onAllLoaded() {
 void Location::onGameLoop() {
 	Object::onGameLoop();
 
+	if (_fadeFramesRemaining > 0) {
+		_fadeFramesRemaining--;
+
+		float newFadeLevel = StarkScene->getFadeLevel() + _fadeLevelIncrement;
+		if (newFadeLevel < 1.0) {
```
bgK: This can be changed to `newFadeLevel = CLIP(StarkScene->getFadeLevel() + _fadeLevelIncrement, 0.0, 1.0);`.

engines/stark/stark.cpp
```diff
@@ -175,6 +175,8 @@ void StarkEngine::setStartupLocation() {
 }
 
 void StarkEngine::mainLoop() {
+	// Location change scripts require >0 loop duration
+	_global->setMillisecondsPerGameloop(10);
```
bgK: Please don't hardcode the frame duration to 10 ms. If you need to make sure it's not zero, add a check in `StarkEngine::updateDisplayScene`.

engines/stark/ui/gamewindow.cpp
```diff
@@ -68,6 +83,14 @@ void GameWindow::onRender() {
 		// Go for the next one
 		element++;
 	}
+
+	float fadeLevel = StarkScene->getFadeLevel();
```
bgK: Please don't do the rendering if there is no fading happening.

engines/stark/ui/gamewindow.cpp
```diff
@@ -51,6 +53,19 @@ GameWindow::GameWindow(Gfx::Driver *gfx, Cursor *cursor, ActionMenu *actionMenu,
 	_position = Common::Rect(Gfx::Driver::kGameViewportWidth, Gfx::Driver::kGameViewportHeight);
 	_position.translate(0, Gfx::Driver::kTopBorderHeight);
 	_visible = true;
+
+
+	_fadeSurface = new Graphics::Surface();
```
bgK: I would prefer if the black rect was rendered using a non textured rectangle, or at least using a single pixel texture. So that we don't need to keep a full size surface + texture in memory and upload it to GPU memory on each frame.

engines/stark/stark.cpp
firesock: The first command in the location change scripts executed by `ResourceProvider::performLocationChange` in the House of the Worlds scene, is an opFadeScene which requires non-zero. Adding a check before doing the location change just causes it to fail during rendering instead. 

Not sure how to have a value for this without rendering a location, do you have any suggestions on this?

bgK: Set a (small) default value from the frame duration when the computed value is zero?

firesock:
Made the requested changes, went with a whole new shader to render a non-textured fade

engines/stark/ui/gamewindow.cpp
```diff
@@ -68,6 +76,11 @@ void GameWindow::onRender() {
 		// Go for the next one
 		element++;
 	}
+
+	float fadeLevel = StarkScene->getFadeLevel();
+	if (fadeLevel < 1.0) {
+		_fadeRenderer->render(fadeLevel);
```
bgK: Looks like some floating point error prevents this check from working as intended when fading is complete:
image

firesock:
Added a quick epsilon diff test, using an arbitrary number

PR #1327, opened 2017-04-02
BUILD: Add back icon for MacOS X bundle

criezy:
This was removed, I assume by mistake, in commit 9ba9c69b that sync'ed
the code with ScummVM.


bgK:
Thanks

PR #1331, opened 2017-04-19
ANDROID: Apply a few missed backports

vpelletier:
These make ResidualVM crash on application initialisation.

Note: this alone is merely enough to get residualvm to start. There are many issues after that:
- adding a game, as of android 7, is impossible as / is "empty", so once there user cannot go down in filesystem tree. Hacked to force /storage/ initial path (not in this PR).
- upon adding a game it crashes because configuration parser somehow finds `vsync=""`, and the emty string is not a boolean (indeed). Hacked parser to return `true` when failing to parse (not in this PR)
- then, shaders are not easily found: they are not copied to the install tree (and makefile does not update that part of the install tree for some reason), and even when shaders are present they are still not found on the device (not sure what made this work, between deleting the install tree, removing the conditional in makefile and requesting the engine static)
- then, before entering game world, debugger window pops up as sound effect rate is not supported (and I am stuck there at the moment)

Apparently, scummvm is also in a sorry state on android. As I have no experience developing on android, I don't think I'll spend much more time on this.

### DIFF
--- a/backends/platform/android/org/residualvm/residualvm/ResidualVM.java
+++ b/backends/platform/android/org/residualvm/residualvm/ResidualVM.java
@@ -53,6 +53,8 @@ public abstract class ResidualVM implements SurfaceHolder.Callback, Runnable {
 	// Callbacks from C++ peer instance
 	abstract protected void getDPI(float[] values);
 	abstract protected void displayMessageOnOSD(String msg);
+	abstract protected void openUrl(String url);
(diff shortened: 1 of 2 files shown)


vpelletier:
I should have mentioned: these backports were likely missed because they apply to `org/scummvm/scummvm/ScummVM*.java`, while this tree is in `org/residualvm/residualvm/ResidualVM*.java`.

PR #1338, opened 2017-05-02
ALL: Add a --bpp option (and simple command-line patches)

vpelletier:
Add a --bpp option (0, 16, 32).
Document existing --engine-speed option.
Drop duplicate default option declarations.

Questions for review:
- is "ALL" prefix good here ?
- any idea why command line options only seem to work (from engine point of view) when *not* showing the launcher (ie, starting directly to game is fine) ?
- is it fine to discard duplicate default option declarations ?

/cc @bgK @raziel-

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -87,7 +87,7 @@ void SurfaceSdlGraphicsManager::setupScreen(uint gameWidth, uint gameHeight, boo
 	if (_fullscreen)
 		sdlflags |= SDL_FULLSCREEN;
 
-	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, 0, sdlflags);
+	_screen = SDL_SetVideoMode(effectiveWidth, effectiveHeight, ConfMan.getInt("bpp"), sdlflags);
 	if (!_screen) {
 		warning("Error: %s", SDL_GetError());
 		g_system->quit();
@@ -381,6 +381,8 @@ SDL_Surface *SurfaceSdlGraphicsManager::SDL_SetVideoMode(int width, int height, 
 	deinitializeRenderer();
 
 	uint32 createWindowFlags = 0;
+	Uint32 rmask, gmask, bmask, amask, format;
+	int depth;
 #ifdef USE_SDL_RESIZABLE_WINDOW
 	createWindowFlags |= SDL_WINDOW_RESIZABLE;
 #endif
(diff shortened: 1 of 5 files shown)


aquadran:
👍 

@bgK any comment on SDL2 part?

raziel-:
@vpelletier 

Getting a compiling error...not sure if i'm allowed to post it here so i'll go with pastebin.
https://pastebin.com/U6ueqHYt

Could very well be a c&p error on my behalf

raziel-:
@vpelletier 

Works great now, thanks a lot

vpelletier:
Fixed a type error in "--bpp" commit (uint32 != Uint32), as noticed by @raziel- .

bgK:
Looks good! Two minor comments:
* If the `--bpp` option only works for the software mode, the documentation should say so (Otherwise, change `OpenGLSdlGraphicsManager::createScreen` to filter the supported pixel formats according to the setting).
* The documentation could also say that the only accepted values are 0, 16 and 32. And/or there could be some user input validation.

vpelletier:
I also realised that I was duplicating "bpp" configuration default right before adding a commit removing similar ones... I'll push a fixed version.

@bgK : My intent was indeed to have it work for opengl too, and looks like I got distracted. And it is resisting: SDL_GL_SetAttribute only specifies a minimal requirement, not the exact value I want, so I end up getting a 32bits GL context when I request for a 16bits one. And from looking at SDL API, there does not seem to be a simple way to get around this ("create window, get surface, change surface bit depth" not being simple to me...). Any better idea ?

vpelletier:
Changes since last push:
- `bpp` documented as only affecting software renderer, to not delay the push as it already improves the situation for @raziel- and makes testing easier
- do not add a _bpp property when it's only used inside a single scope

vpelletier:
I take this opportunity to also add --[no-]dirtyrect option, useful for benchmarking.

Which brings an extra review question: Is it OK to access ConfMan directly from inside TinyGL ? Otherwise I can move it to engine's gfx_tinygl.cpp init call and call `tglEnableDirtyRects` from there.

aquadran:
propably calling directly is not best idea, maybe pass from tinygl init?

vpelletier:
@aquadran : done

bgK:
> My intent was indeed to have it work for opengl too, and looks like I got distracted. And it is resisting: SDL_GL_SetAttribute only specifies a minimal requirement, not the exact value I want, so I end up getting a 32bits GL context when I request for a 16bits one. And from looking at SDL API, there does not seem to be a simple way to get around this ("create window, get surface, change surface bit depth" not being simple to me...). Any better idea ?

If there is no easy way to get `--bpp` working for OpenGL, IMO it's fine to keep the option software only. I can't think of a case where selecting the output pixel format for OpenGL would be useful.

aquadran:
yeah, best to keep option for software renderer. help text should mention this

vpelletier:
Changes since last push:

- drop WIP commit trying to add opengl bpp, as it will not happen (at least in this PR)
- added a warning when bpp has an unexpected value, to make failure cause visible.

PR #1339, opened 2017-05-03
ALL: Improve big-endian support

vpelletier:
I decided to rely more on `PixelBuffer` (and hence on `PixelFormat`) for colorspace conversions.

With these changes, MYST3, GRIM and EMI look correct on residualvm runnning in a powerpc chroot (using qemu-user-static for VM-less emulation) both in 16 and 32 bits modes. ~~MYST is not correct yet.~~

Overall, the rules I decided to follow are:
- When reading from disk, use endian-converting read functions.
- Everywhere else, move `PixelBuffer`s around as they contain their own `PixelFormat`, and pixels are nothing without a format.

This means `SCUMM_BIG_ENDIAN` and `SCUMM_LITTLE_ENDIAN` should ideally not be seen in engines. They do appear in tinygl now because multi-byte opengl texture formats (`INT` & `SHORT`) are endian-based.

Special attention points for the review:
- MYST: ~~I only enabled UNSIGNED_SHORT_5_6_5 because I noticed it was disabled for lack of support. I do not know where such textures appears in-game~~ (found: movie subtitles). ~~On big-endian, the menu and intro video look fine, but in-game is mixed up. I did not do much because I am much less familiar with this engine than GRIM, and because MYST is extremely slow in my powerpc chroot for some reason (like 1 fps).~~ (MYST fixed)
- MYST: save games were created with endian-dependent screenshot thumbnail. I consider little-endian to be the only save games ever actually created. This way savegames should be portable. At least I confirm their screenshot is :) .
- GRAPHICS: I apply several changes directly in /graphics folder. Do these apply to scummvm ? Should I submit there first ?
- TINYGL textures: Now creating a texture is likely slower as resizing code became more flexible. But I think this is not an issue as code is much simpler, avoid pixel format issues, and creating textures *should* be a rare-ish operation (am I right on this assumption ?).
- TINYGL: "components" is now an unused local, which causes a warning. I feel this is legitimate: TinyGL should really take this into account. Should I silence this warning (ex: by just doing some checks on the value, as was done before) ?

/cc @bgK for myst change and @raziel- for native PPC testing

### DIFF
--- a/engines/grim/bitmap.cpp
+++ b/engines/grim/bitmap.cpp
@@ -479,10 +479,7 @@ void BitmapData::convertToColorFormat(int num, const Graphics::PixelFormat &form
 	}
 
 	Graphics::PixelBuffer dst(format, _width * _height, DisposeAfterUse::NO);
-
-	for (int i = 0; i < _width * _height; ++i) {
-		dst.setPixelAt(i, _data[num]);
-	}
+	dst.copyBuffer(0, _width * _height, _data[num]);
 	_data[num].free();
 	_data[num] = dst;
 }
--- a/engines/grim/bitmap.cpp
+++ b/engines/grim/bitmap.cpp
@@ -335,13 +335,12 @@ bool BitmapData::loadTile(Common::SeekableReadStream *o) {
 				uint8 r = (p & 0x001F) << 3;
 				uint8 a = (p & 0x8000) ? 0xFF : 0x00;
 				// Recombine the color components into a 32 bit RGB value
-				uint32 tmp = (r << 24) | (g << 16) | (b << 8) | a;
-				WRITE_BE_UINT32(&d[j], tmp);
+				d[j] = (a << 24) | (b << 16) | (g << 8) | r;
 			}
 		} else if (_bpp == 32) {
 			uint32 *d = (uint32 *)data[i];
 			for (int j = 0; j < _width * _height; ++j) {
-				o->read(&(d[j]), 4);
+				d[j] = o->readUint32LE();
 			}
 		}
 	}
--- a/engines/grim/emi/emi.cpp
+++ b/engines/grim/emi/emi.cpp
@@ -223,7 +223,7 @@ void EMIEngine::storeSaveGameImage(SaveGame *state) {
 	unsigned int texWidth = 256, texHeight = 128;
 	unsigned int size = texWidth * texHeight;
 	Graphics::PixelBuffer buffer = Graphics::PixelBuffer::createBuffer<565>(size, DisposeAfterUse::YES);
-	buffer.clear(texWidth * texHeight);
+	buffer.clear(size);
 	for (unsigned int j = 0; j < 120; j++) {
 		buffer.copyBuffer(j * texWidth, j * width, width, screenshot->getData(0));
 	}
--- a/engines/grim/emi/lua_v2.cpp
+++ b/engines/grim/emi/lua_v2.cpp
@@ -500,25 +500,14 @@ void Lua_V2::ThumbnailFromFile() {
 		delete savedState;
 		return;
 	}
-	uint16 *data = new uint16[dataSize / 2];
+	Graphics::Surface screenshot;
+	screenshot.create(width, height, Graphics::createPixelFormat<565>());
+	uint16 *data = (uint16 *) screenshot.getPixels();
 	for (int l = 0; l < dataSize / 2; l++) {
 		data[l] = savedState->readLEUint16();
 	}
-	Graphics::PixelBuffer buf(Graphics::createPixelFormat<565>(), (byte *)data);
-	Bitmap *screenshot = new Bitmap(buf, width, height, "screenshot");
-	if (!screenshot) {
-		lua_pushnil();
-		warning("Lua_V2::ThumbnailFromFile: Could not restore screenshot from file %s", filename.c_str());
-		delete screenshot;
-		delete[] data;
-		delete savedState;
-		return;
-	}
-
-	screenshot->_data->convertToColorFormat(Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24));
-	g_driver->createSpecialtyTexture(index, screenshot->getData(0).getRawBuffer(), width, height);
-	delete screenshot;
-	delete[] data;
+	g_driver->createSpecialtyTexture(index, screenshot);
+	screenshot.free();
 	savedState->endSection();
 	delete savedState;
 
--- a/engines/grim/gfx_base.cpp
+++ b/engines/grim/gfx_base.cpp
@@ -161,15 +161,14 @@ Math::Matrix4 GfxBase::makeProjMatrix(float fov, float nclip, float fclip) {
 }
 
 
-void GfxBase::createSpecialtyTexture(uint id, uint8 *data, int width, int height) {
-	Graphics::PixelFormat pf(4, 8, 8, 8, 8, 0, 8, 16, 24);
+void GfxBase::createSpecialtyTexture(uint id, const Graphics::Surface &surface) {
 	if (id >= _numSpecialtyTextures)
 		return;
 	if (_specialtyTextures[id]._texture) {
 		destroyTexture(&_specialtyTextures[id]);
 	}
-	_specialtyTextures[id].create(width, height, pf);
-	memcpy(_specialtyTextures[id].getPixels(), data, width * height * pf.bytesPerPixel);
+	_specialtyTextures[id].copyFrom(surface);
+	_specialtyTextures[id].convertToInPlace(Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24));
 	createTexture(&_specialtyTextures[id], nullptr, true);
 }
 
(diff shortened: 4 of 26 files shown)


aquadran:
graphics/pixelformat.cpp is scummvm
graphics/pixelbuffer.cpp is residualvm

graphics/pixelformat.cpp
```diff
@@ -58,7 +58,7 @@ Common::String PixelFormat::toString() const {
 		digits += '0' + 8 - componentLoss;
 	}
 
-	return letters + digits;
+	return letters + digits + '@' + ('0' + bytesPerPixel);
```
aquadran: this change need to go scummvm first

vpelletier: Proposed upstream: scummvm/scummvm#947

vpelletier:
Changes since previous push:
- rebased on latest master, to benefit from `to_float` improvement
- MYST3 is now fixed, though I could not test OpenGL in my chroot

I will try to get opengl working so I can test all engines (and work on TLJ). Not sure it can even work with qemu-user approach (which is otherwise most convenient).

Botje:
I read through the changes; everything makes sense. (except maybe the gl_resizeImage change :P)

vpelletier:
@Botje : Thanks a lot !

About gl_resizeImage, you mean the change is unreadable (and I agree that the diff is hard to read) ? Actually, this change kind of precedes some work I have with texture sampling, but I have big doubts I can get somewhere with decent performance: in my understanding, resampling should happen not when creating the texture, but when rendering it. Which means each drawn textured pixel will do 3 or 4 texture reads, and texture reads are already the hot-spot with a single texture read per pixel... I wonder about using a custom internal texture format where all 4 texels are next to each other, to benefit from CPU cache... Or to to totally thrash that cache and memory, as the texture will be 4 times bigger.

vpelletier:
Something I forgot about this patch stack: I did not test audio at all, as I fail to access my audio devices from inside this chroot (...and I did bind-mount /dev, /dev/pts, /sys and /proc !). For reference: `SDL_AUDIODRIVER=disk SDL_DISKAUDIOFILE=/dev/null`

More about gl_resizeImage*.

First, It is fine for me to drop it from this patch set.

Then, here are the semantic changes:
- replace all `for (int j = 0; j < 3; j++)` loops with `getARGBAt`+`setPixelAt`
- convert `(yi + 1) * xsize_src` into `yi * xsize_src + xsize_src`, allowing to...
- factorise `yi * xsize_src + xi` as `point_offset` for readability (fewer repetitions)
- factorise "can I move down/left" conditions into `space_below` and `space_right` for readability (fewer repetitions)
- move y-dependent values outside of the inner loop: this splits above expression into `point_y_offset` and `point_offset` and requires introducing `curr_yf` so excessive x+y error on one pixel does not affect later pixels on same row without having to recompute `yf`

And something else this commit does and which I completely forgot about is that it also filters alpha channel, where original code copied nearest pixel's alpha verbatim.

vpelletier:
Changes since last push:
- discard/postpone graphics/pixelformat.cpp change so it does not delay this pull request (it is not essential anyway)
- rebased on #1341
- use "MYST3" as subsystem name, not "MYST"

vpelletier:
For issue tracking, the color format change in MYST3 opengl renderer causes a regression. I'm tracking it down.

About my self-imposed rule of using Graphics::PixelBuffers, I wonder if I should rather use Graphics::Surface... This would avoid often having to pass width & height around at least.

vpelletier:
Changes since last push:
- use Graphics::Surface as base for grim engine Texture class to have access to PixelFormat. This is a large change in LoC terms, but reduces code duplication between opengl, shaders and tinygl by applying color map when loading textures, not when sending them to gfx. (I would like to work more on gfx code deduplication... Maybe once endianness is correct ?)
- Made formatType2PixelFormat code shorter, in preparation of the latest commit in this branch, adding pixelFormat2FormatType (and which is an RFC, nothing uses it as I would have to implement an OpenGL variant, and I have no idea where to do this). Also, I fixed many pixel formats I added, which were garbage.
- Worked further on fixing MYST3, dropping byte-swaps which are now unnecessary - and which should fix opengl issues on big endian and which should now allow merging GLES and non-GLES code paths.

vpelletier:
Changes since last push:
- fix OpenGLES build by removing reliance on GL_BGR, and instead convert TGAs to RGBA like other textures. Simplifies code. If the added 25% texture memory usage for TGAs is an issue GL_RGB would be a good alternative. It's easy-ish to do now (the only comlication being the `if` branches in all 3 renderers). Also fixes rendering issues introduced in previous push in EMI with opengl & shader renderers (ex: menu was solid white on opengl, solid black with shaders)

vpelletier:
Changes since last push:
- Simplify shaders: `swapRandB` is not needed anymore, and `useVertexAlpha` invariant is now always true for `drawEMIModelFace` (which is not the only place this gets set in).

I'm not sure about `useVertexAlpha` change: I feel that if this is needed it should rely on something independent from alpha channel presence - or maybe on a flag separate from pixel format ?

vpelletier:
Changes since last push:
- Indeed, transparency really needs to be tracked independently from texture format: shader uniform should probably have just used `_hasAlpha` instead of checking pixel format. The effect is especially visible in EMI set `shi`: hard-enabled, the lady's hat and mouth are invisible, hard-disabled boat ripples in the water are solid black, and following `_hasAlpha` works as before this patch stack.
- Also, restore other uses of `_hasAlpha`, as `.aBits()` is always true now.

vpelletier:
Changes since last push:

- drop gl_resizeImage* change, will be in another pull request (uncovered 2 trivial syntax errors in earlier commits, there are fixed now)
- drop RFC patch, I'll see if I can figure how to get a full version before resubmitting it

Some news on my big-endian testing (using PPC as a BE arch as it seems the one with still reasonably-powerful cpus and still good linux support - so long, sparc !):

- qemu userland emulation looks good, but it tests built-in software rendering only (no sound, no hardware-accelerated GL, not even software mesa GL)
- qemu system emulation with virtio graphic card (both with and without passthrough) look bad themselves (ie, I have visible endianness glitches on emulated desktop)
- I bought a PPC machine but... under linux it has an nvidia card, and X freezes with nouveau (although nouveau does not complain much), and under OSX I still do not have all I need to build residualvm from source

vpelletier:
Changes since last push:
- Re-add support for BGR565 texture color format

vpelletier:
Changes since last push:
- Fix memory leak when converting surface color format in GRIM `loadTga` (used in EMI only)

vpelletier:
Changes since last push:
- Fix segfault when re-entering EMI save menu

For some reason, there is now a 1px line of garbage right and below screenshots in EMI save/load menu. Valgrind complains about uninitialised data. Investigating.

vpelletier:
Changes since last push:
- the source of savegame screenshot garbage in EMI is fixed: it was apparently always present, ad was caused by incomplete PixelBuffer clearing. Fix commit added to this PR.
- More `Graphics::Surface` use in GRIM, reducing code duplication a bit

vpelletier:
From tests by @raziel- on real hardware, software rendering looks fine (though slow), but OpenGL does not.

> Overall, the rules I decided to follow are:
> - When reading from disk, use endian-converting read functions.
> - Everywhere else, move `PixelBuffer`s around as they contain their own `PixelFormat`, and pixels are nothing without a format.

I believe these rules are a mistake, because of the OpenGLES constraint: it does not support `GL_UNSIGNED_INT_8_8_8_8`{,`_REV`}, so we have to use `GL_UNSIGNED_BYTE`. Which means endianness "does not matter" for OpenGL. But, confusingly enough, this is exactly where it starts to matter for TinyGL (aka ResidualVM's internal software renderer), because TinyGL reads pixels using multibyte formats (which are by definition endian-sensitive).

So I have to rework this patch stack.

I could reproduce the issue @raziel- encountered in a PPC64 qemu VM with mesa software renderer, so I should now be able to make progress while checking both opengl and software renderers.

vpelletier:
Quick update. This rework will take time, and I'm starting to feel overwhelmed and just totally stuck. Help very welcome.

One thing to know about current state is that TinyGL z-buffer is rather hackish in its current state:
- zbuffer upload to tinygl follow the same path as background images pixel uploads, which is how I got interested in zbuffer to begin with: it contains multibyte values which must be handled differently (endianness-wise) from color pixels, so I'm trying to separate these code paths (ideally, I want to upload z-buffer using `glTexImage2D` with `TGL_DEPTH_COMPONENT`)
- zbuffer is right-handed (0 = far, 0xffff = near), while it is supposed to be left-handed in OpenGL (0 = near, 0xffff = far), which explains an extra `0xffff - ...` in opengl and shader renderers while tinygl has none
- some zbuffer values rely on overflow: tube flag is at 0xffff, which after going through current Z scaling formula is above z-buffer precision (>0xffff). I've been trying to understand where this formula comes from, with no success. I do understand z-buffer is supposed to have more precision the closer to the camera, and less at the back so I understand the need for a non-linear relationship. I also guess, from the value, that zbm files contain linear distance (w buffer ?), and if I understand correctly 0xffff is near clip plane and 0 the far one (I think 0 never really appears).
- near & far are hard-coded when game is grim, because in one set "near" is obviously wrong. For reference:
```
$ grep -h nclip */*.set | sort | uniq -c
      1         nclip           0.000000 # <- bad
    376         nclip           0.010000 # <- hard-coded in set.cpp
      2         nclip           0.100000
      2         nclip           1.000000
      1         nclip           2.750000
      1         nclip           3.000000
      1         nclip           39.950001
$ grep -h fclip */*.set | sort | uniq -c
     16         fclip           10.000000
     13         fclip           100.000000
      1         fclip           1000.000000
      1         fclip           101.000000
      2         fclip           11.100000
      1         fclip           115.000000
      3         fclip           12.000000
      3         fclip           126.420387
      1         fclip           126.420395
      2         fclip           13.058411
      2         fclip           14.000000
      3         fclip           148.731125
      3         fclip           15.000000
      2         fclip           150.000000
      1         fclip           17.000000
      1         fclip           17.500000
      1         fclip           2.000000
     12         fclip           20.000000
     20         fclip           200.000000
      1         fclip           22.000000
      2         fclip           26.500000
      3         fclip           268.673492
      3         fclip           27.500000
      1         fclip           3.000000
     12         fclip           30.000000
      7         fclip           300.000000
      1         fclip           3.250000
    204         fclip           3276.800049 # <- hard-coded in set.cpp
      1         fclip           40.049999
      7         fclip           42.000000
      3         fclip           4.500000
      1         fclip           45.000000
      2         fclip           49.971405
     14         fclip           5.000000
      8         fclip           50.000000
      3         fclip           500.000000
      1         fclip           5.100000
      6         fclip           5.500000
      3         fclip           6.000000
      2         fclip           7.500000
      4         fclip           89.604820
      4         fclip           9.000000
      1         fclip           99.869202
      2         fclip           99.869225
```

I'm severely lacking in graphic coordinate manipulation knowledge, so I fail to understand the formula I even need. Semi-randomly picking stuff out of wikipedia does not help.

aquadran:
no longer needed

PR #1341, opened 2017-05-04
Fix gcc warnings in get_float on big-endian (take 2)

vpelletier:
Also, rename function to be more consistent with others from common/endian.h .


vpelletier:
Changes since last push:

- s/get_float/READ_LE_FLOAT/ in commit message

(no code change)

bgK:
The code looks good to me.

PR #1343, opened 2017-05-12
TINYGL: Resample texture during rendering

vpelletier:
~~Note: based on #1339.~~

Before this change, TINYGL resizes textures when they are uploaded to GL (`tglTexImage2D`) to the init-configured internal square texture size, and applies bilinear filtering at that time.

This has 2 effects:
- memory: most textures are smaller than (currently) requested internal texture size, so they use more memory once sent to TinyGL
- rendering: interpolation happens browsing original pixel grid by walking the resized pixel grid with borders adjusted to line up with the original image. Resulting texture will show re-sampling artifacts when rendered if it is not aligned to screen (...which is mostly visible for high-resolution original images)

Moving bilinear filtering to render time has potentially a large cost: it needs to read 3 texels for each rendered pixel, and reading texture already requires expensive fetches from RAM.

But this cost can be significantly reduced by relying on CPU data cache: each fetch from main memory actually loads several adjacent words in CPU cache, so accessing nearby bytes is cheap.

This is achieved by repacking the original texture, bringing the 4 possible involved texels (of which 3 will be used, choosing which depending on sampling offset) close to each other.

This again has an obvious cost: each texture takes 4 times its original size. But this cost is offset by being working on texture in its original size, saving lots of ram on smaller textures and should allow increasing internal texture resolution with less negative effects than before (but I did not try this much).

# Functional effect

This change allows implementing simply:

- `TGL_NEAREST` texture filtering (although texture sampler is still chosen at texture upload time, not at render time - I did not check what OpenGL is doing, but at least as of current MYST3 and GRIM engines this assumption is correct)
- `TGL_TEXTURE_WRAP_`{`S`,`T`}

# Visual effect
- ship textures in EMI (very visible in first puzzle, more subtle in shipyard)
- myst becomes a lot sharper, and I feel runs even much smoother. As it is uploading textures very often, old TinyGL code was resampling large (640x640) textures to 512x512 a lot, and probably often in vain as their polygon could be fully off-screen.

# Memory usage effect

[EDIT] Measures updated after fixing a per-`tglTexImage2D`-call memory leak (mostly affecting MYST3)

Game | Set | Before | After, `TGL_LINEAR` | After, `TGL_NEAREST`
---- | --- | ------ | ----------------- | ------------------
GRIM | cn | 78 MB | 67 MB | 58 MB
EMI | first puzzle | 91 MB | 91 MB | 81 MB
MYST3 | irrelevant ? | 82 MB | 103 MB | 71 MB

GRIM is an easy win, many textures are smaller than 256x256, so they were made needlessly large before.
EMI first puzzle is a tough one, as the moving ship in the background is a large texture.
Lastly, MYST is full of high-rez textures, larger than its requested 512x512 internal texture size, so it is a loss. But I believe it is a reasonable one compared to the visual and performance improvement.

# Performance effect

Despite the texel repacking trick, this has a mild negative impact on FPS, especially when disabling dirtyrects.

Here are my measurments, with `--engine-speed=0 --no-dirtyrects`

Game | Set | Before | After
---- | --- | ------ | -----
GRIM | mo | 540 fps | 510 fps
GRIM | cn | 185 fps | 160 fps
GRIM | bi | 165 fps | 150 fps
EMI | first puzzle | 110 fps | 78 fps

I believe this performance decrease to be reasonable. But I'm biased :) .

[EDIT] Not based on #1339 anymore

### DIFF
--- a/engines/myst3/gfx_tinygl_texture.cpp
+++ b/engines/myst3/gfx_tinygl_texture.cpp
@@ -35,7 +35,7 @@ TinyGLTexture::TinyGLTexture(const Graphics::Surface *surface) {
 		sourceFormat = TGL_UNSIGNED_INT_8_8_8_8_REV;
 	} else if (format.bytesPerPixel == 2) {
 		internalFormat = TGL_RGB;
-		sourceFormat = TGL_UNSIGNED_BYTE; // UNSIGNED_SHORT_5_6_5 not provided
+		sourceFormat = TGL_UNSIGNED_SHORT_5_6_5;
 	} else
 		error("Unknown pixel format");
 
--- a/graphics/module.mk
+++ b/graphics/module.mk
@@ -54,6 +54,7 @@ MODULE_OBJS := \
 	tinygl/select.o \
 	tinygl/specbuf.o \
 	tinygl/texture.o \
+	tinygl/texelbuffer.o \
 	tinygl/vertex.o \
 	tinygl/zbuffer.o \
 	tinygl/zline.o \
--- a/graphics/pixelbuffer.cpp
+++ b/graphics/pixelbuffer.cpp
@@ -87,7 +87,17 @@ void PixelBuffer::clear(int length) {
 }
 
 void PixelBuffer::copyBuffer(int thisFrom, int otherFrom, int length, const PixelBuffer &buf) {
-	if (buf._format == _format) {
+	if (buf._format.bytesPerPixel == _format.bytesPerPixel &&
+	    buf._format.rShift == _format.rShift &&
+	    buf._format.gShift == _format.gShift &&
+	    buf._format.bShift == _format.bShift &&
+	    buf._format.rLoss == _format.rLoss &&
+	    buf._format.gLoss == _format.gLoss &&
+	    buf._format.bLoss == _format.bLoss && (
+		_format.aLoss = 8 ||
+		buf._format.aLoss == _format.aLoss
+	    )
+	) {
 		memcpy(_buffer + thisFrom * _format.bytesPerPixel, buf._buffer + otherFrom * _format.bytesPerPixel, length * _format.bytesPerPixel);
 	} else {
 		uint8 r, g, b, a;
--- a/graphics/pixelbuffer.h
+++ b/graphics/pixelbuffer.h
@@ -124,17 +124,22 @@ public:
 	inline void setPixelAt(int pixel, uint32 value) {
 		switch (_format.bytesPerPixel) {
 		case 2:
-			((uint16 *) _buffer)[pixel] = TO_LE_16((uint16) value);
+			((uint16 *) _buffer)[pixel] = value;
 			return;
 		case 3:
 			pixel *= 3;
-			value = TO_LE_32(value);
+#if defined(SCUMM_BIG_ENDIAN)
+			_buffer[pixel + 0] = (value >> 16) & 0xFF;
+			_buffer[pixel + 1] = (value >> 8) & 0xFF;
+			_buffer[pixel + 2] = value & 0xFF;
+#elif defined(SCUMM_LITTLE_ENDIAN)
 			_buffer[pixel + 0] = value & 0xFF;
 			_buffer[pixel + 1] = (value >> 8) & 0xFF;
 			_buffer[pixel + 2] = (value >> 16) & 0xFF;
+#endif
 			return;
 		case 4:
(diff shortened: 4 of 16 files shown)


vpelletier:
MYST opening scene

Before (ignore the black dot in the sky, it's the bird moving around):
image

After:
image

vpelletier:
Changes since last push:
- rebase on latest 1339
- fix per-`tglTexImage2D`-call memory leak

bgK:
The visual change is impressive :sunglasses: 

I did a quick performance test with Myst III (LEIS 3):
* master: 41 fps
* tinygl_texture_pixelformat (#1339): 40 fps
* tinygl_texture_sampler (#1343): 30 fps

vpelletier:
Aha, so it's actually slower - and quite a bit.

How did you measure ? I did not see code for it in MYST3. I'ts very helpful (for me) in GRIM, so maybe it should be controlled by --fps in MYST3 too.

bgK:
TinyGL gets patched through OpenGL by SDL2 on my system. I use MESA's features to measure the FPS:
VBLANK_MODE=0 LIBGL_SHOW_FPS=1 ./residualvm

vpelletier:
> I use MESA's features to measure the FPS:

Nvidia blob here, so no luck.

vpelletier:
This PR introduces a one-pixel offset (or scaling error ?) on EMI save-game screenshots.

These are a bit special: the texture is much larger than the displayed portion and is bordered by solid color (apparently white for the "broken save" and "empty slot" images, black for actual screenshots).

vpelletier:
Changes since last push:
- made independent from #1339, which forces disabling endian-dependent pixel formats
- fixed a typo in `#if defined(TINYGL_LITLE_ENDIAN)` (was missing one `T`), although it is now disabled (see above)

I would be interested by:
- opinion whether this should go in, feature-wise, despite the performance impact (or maybe I should make TinyGL just not request linear filtering ?)
- opinion on the naming of the new file and classes in tinygl
- tests that OpenGL does not regress following this patch set (which should only be related to 	`GRAPHICS: Store pixels in native order in PixelBuffer`), especially on big-endian as I still fail to get reliable results even with linux on native hardware (I should set Xcode up and build residualvm there eventually). I'm not interested in stuff that was and is still broken, although stuff that would not be broken the same way could be interesting.

bgK:
* Regarding `GRAPHICS: Store pixels in native order in PixelBuffer`, I very much agree PixelBuffer should not have to deal with endianness. I have no means to test the changes on big endian though.
* Regarding Myst III: The textures for the sides of the cube are 640x640. They are fully updated by the water effect. The other texture updates (mostly spot items) only update the affected portion of the cube sides.

I did not look at the code in too much detail since I only have a very shallow understanding of how TinyGL works, but the changes look to me good overall.

vpelletier:
Thanks for the review.

For Myst III, indeed water effect costs a lot. I noticed this when facing the "rainbow" door (a handy color format test :) ) on Jnanin: when the door is on the left border of the screen, framerate goes up a lot as we are likely facing a no-sea face, while all 3 other faces have waves.

Also, I just realised wave effect can be disabled, so I think this should not be a big blocker.

Profiling some more, I find that, when using TGL_NEAREST, the hot-spot (30% of cpu time) is PixelBuffer's color format conversion: my framebuffer is (from largest LE shift to lowest, IOW as printed by `ColorFormat.asText`): `_RGB` (`_` because the format is 4 bytes per pixel, despite having no alpha channel), but MYST3 passes textures as `ABGR`, so byte-swapping has to happen.

Tweaking texture formats so no byte-swapping is needed, and making PixelBuffer ignore component difference when going from 1+ bits to 0 bits at same shift accelerates significantly.

But I believe this is a separate improvement, which may need graphics layer to expose its preferred pixel format to engines, and then engines to use it (...or not, if not hot-spot, like myst3 menu textures) when converting textures on load.

So I am very tempted to merge this PR.

vpelletier:
Rebased on latest master, plus 2 commits:
- I used convoluted fixed-point code to try to avoid floating-point operations for coordinates conversion in texture sampler. But it is just slow, and trivial code using floating-point is faster (at least on my machine). I add it in a separate commit, but I am tempted to squash in the commit adding texture sampler.
- I also make the change to PixelBuffer format conversion I just talked about: it will try harder to use the fast memcpy path. No change to engines, so for example Myst3 will not take advantage of this on my machine.

aquadran:
merged on scummvm side

PR #1352, opened 2017-07-26
GRIM: Fix Velasco's reminiscences cutting off. Fixes #356

### LINKED ISSUE #356: GRIM: Velasco's reminiscences are cut off
"The SS Lamancha was her name but don't make me talk about her, cause it... it hurts me..." is the full text, but the audio cuts off after the start of "cause".

ToontownLittleCat:
This adds an additional ```wait_for_message()``` line in Velasco's dialog (dlg_velasco.lua) where he was finishing up talking about his "SS Lamancha" after retrieving his log book from him.

This is done because there's a rare occurrence in the ```re.fake_conversation``` function (which starts alongside the final line) where the sleep timers might finish a bit earlier before he's finished talking, therefore, calling ```velasco:shut_up()``` and cuts him off somewhere in the middle or near the end.  So, this patch removes that possibility and instead starts the said script when he is completely done.

Also, this is not an engine bug, as this also occurs in the original, and possibly the remastered engines.


YakBizzarro:
Sorry for the very late reply, I just seen this pull request :)
The patch is good and works properly. However, when you recompile a modified script with luac, you have to use the option -b. In this way, you can use the original script as "template", in order to create a modified version as close as possible to original and minimize the diff. My bad that this passage is not documented...

I fixed your patch in https://github.com/YakBizzarro/residual/commit/e1ca7fb5e04750aa1bbb7a60f8da589186a10eeb 

Fine for me to merge

bgK:
Thank, I merged @YakBizzarro's version

PR #1356, opened 2017-08-28
SDL: Only recreate SDL2 window when necessary

bgK:
This pull request imports the changes from https://github.com/scummvm/scummvm/pull/937 and adapts them to ResidualVM.

With these changes and when building with SDL2, ResidualVM will keep the same window when going from the launcher to a game, and when toggling the fullscreen mode. That means the OpenGL context can be kept during those operations, making it easy to implement fullscreen toggling in Myst III and TLJ, and no longer requiring to save / restore the game in Grim.
The window will still be recreated when switching between software and OpenGL.

The notable changes are as follow:
* The default launcher size is now 640x480 instead of 640x400. This avoids resizing the window when starting a game in windowed mode from the launcher.
* The launcher can now be used in fullscreen mode.
* The launcher will now render in OpenGL mode if selected in the option screen. This avoids recreating the window when starting a game from the launcher.
* Myst III and TLJ now support fullscreen toggling (SDL2 only)
* Grim no longer saves / restores when going to / leaving fullscreen (SDL2 only)

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -155,11 +155,7 @@ void OpenGLSdlGraphicsManager::setupScreen(uint gameWidth, uint gameHeight, bool
 	initializeOpenGLContext();
 	_surfaceRenderer = OpenGL::createBestSurfaceRenderer();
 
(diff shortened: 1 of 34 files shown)


aquadran:
👍

PR #1372, opened 2017-11-05
I18N: Update German GUI translations; fixes

rootfather:
In this PR I update the German GUI translation. Additionally, I provide a default residualvm.pot template file to make translations into other languages easier. I removed a leftover reference to the MT32 emulator that made it impossible to rebuild the .po files. The translations.dat file was also updated.

This could also be backported to the 0.3.0 branch.


rootfather:
I just want to bring this back to your attention - this PR should be ready to get merged. The travis-ci warning is not caused by the changes I made; it seems there are issues with the android checks/toolchain.

PR #1380, opened 2017-11-28
GRIM: Cut back on the number of glPushMatrix operations

klusark:
Most graphics cards limit us to a stack size of 32. In some places
like the pigeons on the roof of the DOD, they will render too many
matricies pushed to the stack.

If we don't need to push it to the stack, skip pushing.

Creating this as a PR hoping someone will check my logic.


bgK:
I've read the code, and the changes make sense to me. :+1:

PR #1386, opened 2017-12-13
MYST3: fix Polish version detection and subtitles

dagashiya:
Hi,
I added a checksum for the Polish version of Myst 3 (4 CD Play-It version, it's actually dual-language) and fixed encoding of the subtitles (which also eliminates crashes). This version is not using windows-1250 encoding but rather a MacOS Central European one.

Some images:
[Subtitles before encoding fix](https://gs.smuglo.li/file/49c43f3c615e5117bc72851b6e4b0cbc1c58e2a44e7587ad3bf5cd97c64a3781.png)
[Subtitles leading to crashes](https://gs.smuglo.li/file/9dd6a2f8cee8f2e4761dbede9a85406a5ec23675806c4212174005b0ac9e1c40.png)
[With fixed encoding](https://gs.smuglo.li/file/9efb1940aecf4b3ad5cc815ff9151634a94189711882a1ce2872fcc58eb7e950.png)

### DIFF
--- a/engines/myst3/detection.cpp
+++ b/engines/myst3/detection.cpp
@@ -115,7 +115,7 @@ static const Myst3GameDescription gameDescriptions[] = {
 	MYST3ENTRY(Common::DE_DEU, "GERMAN.m3u",   "1b2fa162a951fa4ed65617dd3f0c8a53", 0, kLocMulti2) // #1323, andrews05
 	MYST3ENTRY(Common::IT_ITA, "ITALIAN.m3u",  "906645a87ac1cbbd2b88c277c2b4fda2", 0, kLocMulti2) // #1323, andrews05
 	MYST3ENTRY(Common::ES_ESP, "SPANISH.m3u",  "28003569d9536cbdf6020aee8e9bcd15", 0, kLocMulti2) // #1323, goodoldgeorge
-	MYST3ENTRY(Common::PL_POL, "POLISH.m3u",   "00000000000000000000000000000000", 0, kLocMulti2)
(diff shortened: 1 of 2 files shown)


bgK:
Thanks for you contribution!

I'm pretty sure the Polish subtitles had the correct encoding at some point before. Would you mind checking your CDs to see if there are different Polish language files on the Windows and Mac partitions?

dagashiya:
The funny thing is, I'm pretty sure it worked before too. I always used Joliet filesystem but the discs do have HFS on them.
I'm currently compiling checksums of all files visible in HFS, then I'll run a check on the disc mounted normally. If data is the same, it should only complain about missing files.

dagashiya:
There are no differences between game data in standard ISO and HFS filesystem except for MacOS-specific files.

Methodology:

- mount CD-ROM block device as HFS
- use find and md5sum tools to make checksums of every file, dump the result to a text file
- remount the disc as iso9660
- check the checksums while dumping the output to another text file
- read the file and filter out every line that shows a file matches (thus leaving only errors and warnings)
- swap to another CD and start again

Here is a [zipfile](https://0x0.st/sXVS.zip) with both checksums from HFS and output of their checks while working on Joliet.

bgK:
Thanks for taking the time to check the files!

dagashiya:
Hi again, there might be a new issue connected to this. [This](https://0x0.st/sXYj.png) is what happens on first Polish subtitle display in the latest daily build of ResidualVM for Windows. An iconv problem?

PR #1392, opened 2018-01-26
STARK: Make mouse hints appear more like in the game

Snejp:
Added screen boundary to mouse hints. Made hints static for topmenu buttons. Proper text placement in fullscreen.

### DIFF
--- a/engines/stark/ui/world/topmenu.cpp
+++ b/engines/stark/ui/world/topmenu.cpp
@@ -42,8 +42,8 @@ TopMenu::TopMenu(Gfx::Driver *gfx, Cursor *cursor) :
 	_visible = true;
 
 	_inventoryButton = new Button("Inventory", StaticProvider::kInventory, Common::Point(32, 2), Button::HintAlign::kLeft, Common::Point(64, 20));
-	_optionsButton = new Button("Options", StaticProvider::kDiaryNormal, Common::Point(560, 2), Button::HintAlign::kRight, Common::Point(523+37, 20));
-	_exitButton = new Button("Quit", StaticProvider::kQuit, Common::Point(608, 2), Button::HintAlign::kRight, Common::Point(587+21, 20));
+	_optionsButton = new Button("Options", StaticProvider::kDiaryNormal, Common::Point(560, 2), Button::HintAlign::kRight, Common::Point(560, 20));
+	_exitButton = new Button("Quit", StaticProvider::kQuit, Common::Point(608, 2), Button::HintAlign::kRight, Common::Point(608, 20));
 }
 
 TopMenu::~TopMenu() {
--- a/engines/stark/ui/world/topmenu.cpp
+++ b/engines/stark/ui/world/topmenu.cpp
@@ -41,9 +41,9 @@ TopMenu::TopMenu(Gfx::Driver *gfx, Cursor *cursor) :
 	_position = Common::Rect(Gfx::Driver::kOriginalWidth, Gfx::Driver::kTopBorderHeight);
 	_visible = true;
 
-	_inventoryButton = new Button("Inventory", StaticProvider::kInventory, Common::Point(32, 2));
-	_optionsButton = new Button("Options", StaticProvider::kDiaryNormal, Common::Point(560, 2));
-	_exitButton = new Button("Quit", StaticProvider::kQuit, Common::Point(608, 2));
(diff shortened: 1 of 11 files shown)


engines/stark/ui/world/topmenu.cpp
```diff
@@ -41,9 +41,9 @@ TopMenu::TopMenu(Gfx::Driver *gfx, Cursor *cursor) :
 	_position = Common::Rect(Gfx::Driver::kOriginalWidth, Gfx::Driver::kTopBorderHeight);
 	_visible = true;
 
-	_inventoryButton = new Button("Inventory", StaticProvider::kInventory, Common::Point(32, 2));
-	_optionsButton = new Button("Options", StaticProvider::kDiaryNormal, Common::Point(560, 2));
-	_exitButton = new Button("Quit", StaticProvider::kQuit, Common::Point(608, 2));
+	_inventoryButton = new Button("Inventory", StaticProvider::kInventory, Common::Point(32, 2), Common::Point(64, 20));
+	_optionsButton = new Button("Options", StaticProvider::kDiaryNormal, Common::Point(560, 2), Common::Point(523, 20));
```
bgK: Using the top left position of the hint for the options and quit buttons is going to be an issue when we support loading the labels from the translation files. The non-english text will not have the same length. The top right position needs to be used.

engines/stark/ui/world/topmenu.cpp
```diff
@@ -69,7 +69,7 @@ void TopMenu::onMouseMove(const Common::Point &pos) {
 		Button *hoveredButton = getButtonAtPosition(pos);
 		if (hoveredButton) {
 			_cursor->setCursorType(Cursor::kActive);
-			_cursor->setMouseHint(hoveredButton->getText());
+			_cursor->setMouseHint(hoveredButton->getText(), hoveredButton->getHintPosition());
```
bgK: I suggest not reusing the same code as for world object mouse hints for the top menu button hints. They don't work the same way. This will make the code in the Cursor class cleaner and make it easier to implement the mouse hint delay.
The code for the top menu hints can be in the TopMenu class or in the Button class.

Snejp:
I corrected the code. I would like too make more changes so could you send me the disassembly you mentioned?

engines/stark/visual/text.cpp
```diff
@@ -109,7 +115,7 @@ void VisualText::createTexture() {
 
 	// Render the lines to the surface
 	for (uint i = 0; i < lines.size(); i++) {
-		font->drawString(&surface, lines[i], 0, scaledLineHeight * i, scaledRect.width(), _color);
+		font->drawString(&surface, lines[i], 0, scaledLineHeight * i - 1, scaledRect.width(), _color);
```
bgK: Are you sure about this change? This means y = -1 for the first line which does not seem correct.

Snejp:
I have switched to scaled mode using `_gfx->setScreenViewport(false);`. Now the hints are no more fluid like they were before. Tried many solutions but in this mode mouse labels are moving/jumping approximately 2-3px in fullscreen. Mouse cursor is too not smooth when it is in `_gfx->setScreenViewport(false);` I noticed that text sometimes is bad scaled too.


I have not completed this commit yet because i need tips for width and height of mouse hints because i deleted `->getTexture()`.
```c++
pos.y = CLIP<int16>(pos.y, _gfx->kTopBorderHeight, _gfx->kOriginalHeight - _gfx->kBottomBorderHeight - cursorDistance /* -_mouseText->actualTextHeight() */ ); 
//pos.x -= _mouseText->actualTextWidth() / 2; 
```
I need the width of the hint texture because the label has too be centered under the cursor and the height to stop it appear in the bottom border(where are subtitles).

```c++
Common::Point pos(_hintPosition);
	if (_align == HintAlign::kRight) {
		//pos.x -= _mouseText->actualTextWidth();
	}
_mouseText->render(pos);
```
Here i need the width of the rendered text label to float it to the right next to buttons.

```c++
font->drawString(&surface, lines[i], 0, scaledLineHeight * i - 1, scaledRect.width(), _color);
```
About that y = -1 you are right. I did this only to myself because text in hints are rendering 1px too low and i didn't know how to do this right and after that change it looked identical like in the game. Here are examples (top is actual game, bottom is residualvm):
original april
residualvm april

bgK:
Looks good!

PR #1395, opened 2018-02-09
STARK: Fix not matching mouse hints

Snejp:
There was not proper text hints in some scenes for objects and background because of wrong index.


bgK:
Looks good, thanks!

PR #1398, opened 2018-02-10
Support fullscreen on RPI (using framebuffer)

lmerckx:
Raspberry PI doesn't support 16 depth framebuffer. This patch uses 24 when capable (16 if not).
This resolve the problem and allow to run ResidualVM fullscreen on RPI platforms (with 3D acceleration).

### DIFF
--- a/graphics/opengl/framebuffer.cpp
+++ b/graphics/opengl/framebuffer.cpp
@@ -49,6 +49,7 @@
 
 #ifdef USE_GLES2
 #define GL_DEPTH24_STENCIL8 GL_DEPTH24_STENCIL8_OES
+#define GL_DEPTH_COMPONENT24 GL_DEPTH_COMPONENT24_OES
 #endif
 
 namespace OpenGL {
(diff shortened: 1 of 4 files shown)


graphics/opengl/framebuffer.cpp
```diff
@@ -154,7 +158,7 @@ void FrameBuffer::init() {
 		glBindRenderbuffer(GL_RENDERBUFFER, 0);
 	} else {
 		glBindRenderbuffer(GL_RENDERBUFFER, _renderBuffers[0]);
-		glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH_COMPONENT16, _texWidth, _texHeight);
+		glRenderbufferStorage(GL_RENDERBUFFER, useDepthComponent24()?GL_DEPTH_COMPONENT24:GL_DEPTH_COMPONENT16, _texWidth, _texHeight);
```
bgK: Please surround the `?` and the `:` with spaces.
See http://wiki.scummvm.org/index.php/Code_Formatting_Conventions

lmerckx: No problem, it's done.

PR #1404, opened 2018-02-20
Improve joystick usage

lmerckx:
Improve joystick usage:
- configuration of engines natively supporting joystick or not
- for first category (GrimE), supports a special button from joystick to open the virtual keyboard,
  then emulates mouse from joystick to select keyboard keys
- for second category (Myst3), reuses ScummVM code to emulate mouse from joystick.
  A modification was also needed to support relative movements (needed in Myst3 to support camera movements)

### DIFF
--- a/backends/events/sdl/resvm-sdl-events.cpp
+++ b/backends/events/sdl/resvm-sdl-events.cpp
@@ -25,24 +25,62 @@
 #if defined(SDL_BACKEND)
 
 #include "resvm-sdl-events.h"
+#include "graphics/cursorman.h"
+#include "engines/engine.h"
+
+// The maximum index of the joystick button to convert in mouse event
+#define JOY_TO_MOUSE_BUTTONS 1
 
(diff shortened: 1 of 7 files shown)


bgK:
Thanks for your contribution. I have pushed a modified version of your changes to master.
* I've altered the code to only change ResidualVM specific files. The core files from ScummVM are periodically merged into ResidualVM. Having small changes in ScummVM files in ResidualVM makes that process harder.
* As a result, I have removed the virtual keyboard changes. Those would apply to ScummVM as well. Maybe you can open a ScummVM pull request with the virtual keyboard changes? That way they would be copied in ResidualVM the next time the projects are synchronized.

lmerckx:
Ok, thank you bgK.
With the new gamecontroller feature of ScummVM, it will be easier, I think.

By the way, do you think it is possible for ScummVM to load a custom game controller mapping file with the function SDL_GameControllerAddMappingsFromFile ?
Some controllers are not supported by default with SDL.

PR #1406, opened 2018-03-11
SDL: Do not try to replace FILE with an incompatible type

rootfather:
This fixes broken compilation on current msys2/mingw-w64 environments.
Originally written by @csnover for the ScummVM project.

csnover says:

"It’s questionable that the SDL backend code is trying so hard to
maintain forbidden symbols at all since backend code is exactly
where such things are allowed, and it is a game of whack-a-mole
to keep system API changes from breaking this fragile system of
symbol redefinitions. Probably this should all just get replaced
with a FORBIDDEN_SYMBOL_ALLOW_ALL, but for the time being this at
least gets builds working again on up-to-date msys2/mingw-w64
compilers."

### DIFF
--- a/backends/platform/sdl/sdl-sys.h
+++ b/backends/platform/sdl/sdl-sys.h
@@ -31,15 +31,8 @@
 
 #include "common/scummsys.h"
 
-// Remove FILE override from common/forbidden.h, and replace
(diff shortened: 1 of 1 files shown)


bgK:
Thanks. A full synchronization with the ScummVM code has been performed. These changes are included.

Issue #1402, opened 2018-02-19
STARK: Diary index text is not properly resized when switching between the window and fullscreen mode


bgK:
In The Longest Journey, while in fullscreen mode, in the diary index, toggle into the windowed mode (alt-enter). The menu items captions are too big:

![screenshot from 2018-02-19 20-42-15](https://user-images.githubusercontent.com/52294/36394604-616333ee-15b5-11e8-8f92-37ae9ad024bc.png)

This is because the text textures were not redrawn to account for the smaller font scale.

To fix the issue, implement a onScreenChanged method in the DiaryIndex class that resets the text texture, like in other user interface elements.

PR #1407, opened 2018-03-23
STARK: Fixing the text resized problem on diary index menu

DouglasLiuGamer:
Fix the issue #1402.

I want to change as less code as possible, which was a little bit hard since all the functions related to resetting text texture are encapsulated, so I did in an indirect way. I "forced" the render entry to reset the texture by changing the text color to another and then change it back. Also I had to add a function on StaticLocationWidget to check whether it is safe to call the setTextColor function.

Honestly all these are not that ideal from my perspective. It would be a lot more elegant if I can add some other functions or use friend function, but I am afraid that this will violate the original encapsulation.

### DIFF
--- a/engines/stark/ui/menu/diaryindex.cpp
+++ b/engines/stark/ui/menu/diaryindex.cpp
@@ -102,9 +102,9 @@ void DiaryIndexScreen::open() {
 	}
 }
 
-// Douglas' change
 void DiaryIndexScreen::onScreenChanged() {
 	for (uint i = 1; i < _widgets.size(); i++) {
+		// Force the texture to be reset by changing the color
 		if (_widgets[i]->containsText()) {
 			_widgets[i]->setTextColor(_textColorHovered);
 			_widgets[i]->setTextColor(_textColorDefault);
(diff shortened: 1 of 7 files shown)


engines/stark/ui/menu/diaryindex.cpp
```diff
@@ -102,6 +102,16 @@ void DiaryIndexScreen::open() {
 	}
 }
 
+void DiaryIndexScreen::onScreenChanged() {
+	for (uint i = 1; i < _widgets.size(); i++) {
+		// Force the texture to be reset by changing the color
+		if (_widgets[i]->containsText()) {
+			_widgets[i]->setTextColor(_textColorHovered);
+			_widgets[i]->setTextColor(_textColorDefault);
```
bgK: Here, calling `setTextColor` twice to reset the texture, while a perfectly working solution feels like a hack. Now when changing `setTextColor`, we need to make sure it does not break screen resizing. It would be better to have a specialized call to reset the text visual. One that could be reused to reset other text visuals in the game where this problem occurs.

DouglasLiuGamer: Right this puts an extra dependency on ```setTextColor```... As said in the description, I don’t want to change too many codes, but now I guess opening a specialized interface on class ```VisualText``` is likely a must. I’ll commit this as soon as possible.

bgK:
Very good, thanks!

Issue #1401, opened 2018-02-19
STARK: Sticky black borders in full screen mode


bgK:
When playing The Longest Journey in full screen mode on a wide screen, black borders are drawn on the sides of the screen to preserve the original aspect ratio.

The issue is the mouse cursor feels like it sticks to the black borders. This is because the mouse cursor is hidden when the mouse is over the borders.

To fix the issue, the mouse cursor needs to be made visible when over the black borders. The cursor image needs to be that of the inactive game cursor.

This bug appears everywhere in the game. The demo version of the game can be used: https://archive.org/details/Longest_Journey_demo

PR #1408, opened 2018-03-26
STARK: Fix for the cursor over border issue #1401

mj0331:
Removing the coordinate bounds will make the cursor show up over the black border, as requested in issue #1401 , however it's possible to interact with elements 'hidden' by the border(for example moving the character 'offscreen'). A possible improvement/fix would be to have a bounded cursor position for game logic and a unbounded position for rendering


mj0331:
As a sidenote, I know it would have been better to start a new branch instead of working on master, although I was a bit pressured by time. I'll start new branches in the future

engines/stark/ui/cursor.cpp
```diff
@@ -74,7 +74,7 @@ void Cursor::setCursorImage(VisualImageXMG *image) {
 
 
 void Cursor::setMousePosition(const Common::Point &pos) {
-	_mousePos = _gfx->getScreenPosBounded(pos);
```
bgK: Ok, now `getScreenPosBounded` is not used anymore and can be removed.

Also this does not make sure the cursor is set to inactive when over the black borders. For example, in the TLJ demo, when hovering the quit button in the top bar and moving the mouse to the right over the black border, the cursor stays active.

mj0331: Would adding a function `Gfx::Driver::isPosInScreenBounds` be fine? I could probably use that to do some checks for borders, maybe in `Stark::Cursor::Render`

bgK: Ok, sounds fine. Did you notice `Driver::getScreenViewport`?

mj0331: Yeah, shall I use that for bound checking?

bgK: You might want to.

mj0331:
I think I need to scale the cursor's position when the game is in fullscreen, because just a direct `_screenViewport.contains(pos)` doesn't seem to cut it. I'm not 100% sure what and how to scale it so I'm comparing things in the same coord space

bgK:
Cursor::render already manipulates both original and native coordinates. Does it not work when making the test using native coordinates?

mj0331:
With both native and original coordinates, the bound test fails. My render function looks like this, but it thinks the cursor is out of bounds when it's on the left half of the screen:
https://gist.github.com/mj0331/1f61944594da69542693d0baeb837c84

bgK:
`isPosInScreenBounds` expects native coordinates while you are passing it original resolution coordinates. Using native mouse coordinates should work:
```cpp
    if (!_gfx->isPosInScreenBounds(_mousePos)) {
        setCursorType(Cursor::kPassive);
    }
```

mj0331:
Yeah, I tried with native coords as well, and the bound check still fails when the cursor is in the left quarter of the screen. I'll need to debug this a little further

mj0331:
Alright, this should be it

bgK:
> Alright, this should be it

Yes, looks ok. One last little thing. Notice how your code uses spaces for intentation while the rest uses tabs. Please change that.
(These are the code formatting conventions we follow: http://wiki.scummvm.org/index.php/Code_Formatting_Conventions)

mj0331:
Oops, I thought CLion converted those into tabs but I was wrong. My bad

bgK:
All good, thanks!

PR #1410, opened 2018-04-07
STARK: Add right-click dialog speech skipping

Snejp:
You can skip speech lines talked by game characters showed in dialogPanel with a mouse right-click


bgK:
Thanks for your contribution!

PR #1415, opened 2018-05-15
STARK: Implement the main menu

DouglasLiuGamer:
This pull request is corresponding to the task of implementing the main menu in TLJ.

The first attempt I just did some simple things.

I move ```onScreenChanged()``` to the ```StaticLocationScreen``` since I think all of the menus will need to resize text.

Now I just make a simple ```MainMenuScreen``` class with nothing in it and let the game load it when it starts, so that I can use the debug mode to dump the resources. But now there is a **major problem**. When I do this, an assertion failed in resourcereference.cpp, line 61, as below
```
assert(resource->getIndex() == element.getIndex());
```
I printed out those values. Most of the time both of them are zeros but at a point ```element.getIndex()``` becomes 69.

I guess this is related to how TLG loads its resources and now I am still not very clear about it.

FYI, I add the line ```_currentScreen->open()``` in ```UserInterface::init()```, which I think is necessary. The reason why it wasn't there before, I guess, is that the game will trigger a video to play when it starts, which triggers the```changeScreen()``` and therefore trigger the ```open()```.

~~I made some stupid things in sync my branch and a typo in the name of commits. Hope it is fine.~~


DouglasLiuGamer:
I have rebased my branches to clean up the mess I left before now. Thanks for the help 😄 

Right now the game shall open with the main menu. But now there is another **major problem**: For some reason, all buttons on the main menu separate the clicking area and the helping text. Due to this reason, the original ```isMouseInside()``` is not able to correctly check the bounding area since it just check ```image```, ```smacker``` and ```text```, while those buttons' areas contain only ```text``` and are very tiny. Right now I am not sure how to get the original bounding area. I tried to compare the resources dumped from main menu and diary index but there is nothing remarkable.

There is another small problem: there are two special widgets on the main menu: "VERSION INFO" and "VERSION INFO REALLY". Strange names and the current version renders them differently compared to the steam version, as below:

Steam version:
steam

ResidualVM version:
vm

Seems like those texts are generated by other things. Not sure about this.

engines/stark/visual/text.cpp
```diff
@@ -144,4 +159,13 @@ void VisualText::resetTexture() {
 	createTexture();
 }
 
+bool VisualText::isBlank() {
+	for (auto c : _text) {
+		if (c != ' ' && c != '\r' && c != '\n' && c != '\t') {
```
bgK: You can use `isSpace` from `util.h` instead.

engines/stark/visual/text.cpp
```diff
@@ -144,4 +159,13 @@ void VisualText::resetTexture() {
 	createTexture();
 }
 
+bool VisualText::isBlank() {
+	for (auto c : _text) {
```
bgK: We can't use c++11 features yet. That means no foreach and no auto :(

ResidualVM follows the same rules as ScummVM, and we don't do c++11 there either. This is mostly because the CI server has not yet been updated with c++11 compatible compilers for all platforms.

engines/stark/visual/text.h
```diff
@@ -62,6 +63,9 @@ class VisualText : public Visual {
 	void createTexture();
 	void freeTexture();
 
+	// Check whether the text is blank
```
bgK: Please use Doxygen style comments when documenting methods/classes (so they appear in http://doxygen.residualvm.org/).

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -0,0 +1,104 @@
...
+    _widgets.push_back(new StaticLocationWidget(
+			"BGImage",
+			nullptr,
+			nullptr));
+    
+    _widgets.push_back(new StaticLocationWidget(
+			"NewGame",
+			nullptr,
+			nullptr));
+	_widgets.back()->setupSounds(0, 1);
```
bgK: Please fix the whitespaces here. We use tabs for indentation and spaces for alignment (see http://wiki.scummvm.org/index.php/Code_Formatting_Conventions).

engines/stark/visual/text.cpp
```diff
@@ -124,8 +124,8 @@ void VisualText::createTexture() {
 		_originalRect.bottom = _originalRect.top + originalLineHeight * lines.size();
 	} else {
 		// For Empty text, preserve the original width and height for being used as clicking area
-		_originalRect.right = _originalRect.left + maxScaledLineWidth;
-		_originalRect.bottom = _originalRect.top + StarkGfx->scaleHeightOriginalToCurrent(_targetHeight);
+		_originalRect.right = _originalRect.left + _targetWidth;
+		_originalRect.bottom = _originalRect.top + _targetHeight;
```
DouglasLiuGamer: The previous implementation is wrong. The width and height should not be scaled according to the resolution since the testing of the bounding box is based on original resolution.

engines/stark/visual/text.cpp
DouglasLiuGamer: I see that the ```override``` specifier is used heavily so I thought c++11 is available. Looks like the c++11-compat.h redefines it. Neat.

bgK:
This is working quite well now :+1:

engines/stark/services/userinterface.cpp
```diff
@@ -67,9 +69,14 @@ UserInterface::~UserInterface() {
 void UserInterface::init() {
 	_cursor = new Cursor(_gfx);
 
-	_currentScreen = _gameScreen = new GameScreen(_gfx, _cursor);
+	_mainMenuScreen = new MainMenuScreen(_gfx, _cursor);
```
bgK: `_mainMenuScreen` is never freed (memory leak).

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -0,0 +1,111 @@
+#include "engines/stark/ui/menu/mainmenu.h"
```
bgK: Please include the standard ResidualVM licence header for new files.

engines/stark/services/services.h
```diff
@@ -97,7 +97,7 @@ class StarkServices : public Common::Singleton<StarkServices> {
 #define StarkGameInterface      StarkServices::instance().gameInterface
 #define StarkUserInterface      StarkServices::instance().userInterface
 #define StarkFontProvider       StarkServices::instance().fontProvider
-#define StarkGameDescription	StarkServices::instance().gameDescription
+#define StarkGameDescription    StarkServices::instance().gameDescription
```
DouglasLiuGamer: Add this field so that MainMenuScreen may check the game version.

engines/stark/ui/menu/locationscreen.cpp
```diff
@@ -180,6 +180,9 @@ void StaticLocationWidget::onClick() {
 		_soundMouseClick->play();
 	}
 
+	// Ensure the click sound is played completely
+	while (_soundMouseClick->isPlaying()) {};
+
```
DouglasLiuGamer: Based on the original game (steam version here), the click sound is completely played before the action is executed. Although using ```while``` loop here will freeze the game, but that is also what will happen in the original game, so this should be enough "original...ish".

engines/stark/services/userinterface.cpp
DouglasLiuGamer: Fixed.

DouglasLiuGamer:
I am planning to implement the Quit and Credits, but I am not sure how to find out their corresponding resources.

For Credits I need to load the credits fmv but right now I have no idea how to find the name. As for Quit, according to the original game (steam version), when the user hits the button, there will be a message box coming up to let user chooses "yes" or "no" with either mouse or keyboard, no idea how to find those resources. I can't screenshot the message box since the steam version has problems rendering it properly.

@bgK Maybe we can discuss on email if the solution is a bit complex.

bgK:
I don't have enough time to write a full answer right now. The credits video is `global\xarc\0e02.bbb`. The Funcom logo that plays before the main menu is displayed is `global\xarc\1402.bbb`. Both are hardcoded in the game executable. I don't think there are resource descriptors.

There are a few other confirmation dialog boxes. I believe those are actually skinned Windows dialogs. We'll have to come up with a replacement solution. Perhaps we could use ResidualVM GUI dialogs.

engines/stark/ui/menu/locationscreen.cpp
DouglasLiuGamer: Wait the pointer needs to be checked beforehand.

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -132,6 +132,12 @@ void MainMenuScreen::helpTextHandler(StaticLocationWidget &widget, const Common:
 	}
 }
 
+void MainMenuScreen::creditsHandler() {
+	if (!isDemo()) {
+		StarkUserInterface->requestFMVPlayback("0e02.bbb");
```
DouglasLiuGamer: There is no 0e02.bbb in the demo game data. So demo has no credits video or with a different name?

DouglasLiuGamer:
One small ~~and strange yet a little bit funny~~ thing to ask: in my steam version game, before the FunCom logo video, there is a screen that fades very quickly with a text "Please Wait...". Do we need to make that too? 😂😂😂

engines/stark/ui/menu/locationscreen.cpp
```diff
@@ -171,6 +178,8 @@ void StaticLocationWidget::onClick() {
 
 	if (_soundMouseClick) {
 		_soundMouseClick->play();
+		// Ensure the click sound is played completely
+		while (_soundMouseClick->isPlaying()) {};
```
bgK: It's not good to busy-wait for something to complete. The game appears unresponsive while waiting and the CPU is used at 100%.
In game engine design it's best to have a single game loop that roughly checks for user input, updates the simulated world and then renders. So far, Stark has managed to follow that rule. So here, we'll need to find another solution to let the sound fully play. Perhaps something is stopping the sound, preventing it from completing while it should not. Perhaps this code needs to wait for the sound to complete by checking each frame if it is finished.

engines/stark/services/userinterface.cpp
```diff
@@ -134,24 +138,26 @@ void UserInterface::requestFMVPlayback(const Common::String &name) {
 }
 
 void UserInterface::onFMVStopped() {
-	backPreScreen();
+	backPrevScreen();
```
DouglasLiuGamer: This commit contains some changing of names like this.  Just my own decision. Nothing remarkable.

DouglasLiuGamer: Just think these names aren't appropriate.

engines/stark/services/userinterface.cpp
```diff
@@ -264,9 +264,11 @@ const Graphics::Surface *UserInterface::getGameWindowThumbnail() const {
 }
 
 void UserInterface::onScreenChanged() {
-	_mainMenuScreen->onScreenChanged();
 	_gameScreen->onScreenChanged();
-	_diaryIndexScreen->onScreenChanged();
+
+	if (!isInGameScreen()) {
+		_currentScreen->onScreenChanged();
```
DouglasLiuGamer: There will be more ```Screen``` added to ```UserInterface``` later. Since ```StaticLocationScreen``` will release all the resources when it is closed, there is no need to notify the change of screen resolution for a not currently displayed screen.

DouglasLiuGamer:
**One problem**: even though now the game will "quit" to the main menu, if choose to start the game again, the game will just go back to the previous state, not start a completely new one. Is there a way to reset the whole ```GameScreen```, or maybe something to twist in ```ResourceProvider```?

engines/stark/services/userinterface.cpp
```diff
@@ -158,6 +161,13 @@ void UserInterface::backPrevScreen() {
 	changeScreen(_prevScreenNameStack.pop());
 }
 
+void UserInterface::quitToMainMenu() {
+	changeScreen(Screen::kScreenGame);
+	StarkResourceProvider->shutdown();
+	StarkResourceProvider->initGlobal();
+	changeScreen(Screen::kScreenMainMenu);
```
DouglasLiuGamer: It feels like here should have a ```StarkResourceProvider->setShouldRestoreCurrentState();```, but that would corrupt the game when it starts again. Kind of want to know why.

bgK: It would make more sense if `StarkResourceProvider->initGlobal();` was not called here (nor on engine startup when going to the main menu), but when starting a new game.

`StarkResourceProvider->setShouldRestoreCurrentState();` is part of restoring a saved game. It asks the resource provider to restore the game state part that is specific to the current location. There's no need to call it here nor when starting a new game.

engines/stark/stark.cpp
```diff
@@ -138,22 +138,16 @@ Common::Error StarkEngine::run() {
...
 	_resourceProvider->initGlobal();
 	_staticProvider->init();
 	_fontProvider->initFonts();
+	
 	// Initialize the UI
 	_userInterface->init();
 
-	if (ConfMan.hasKey("save_slot")) {
-		// Load game from specified slot, if any
-		loadGameState(ConfMan.getInt("save_slot"));
```
bgK: Ultimately, it should still be possible to load a save from the ResidualVM launcher whithout going through the main menu.

engines/stark/ui/menu/locationscreen.cpp
```diff
@@ -171,6 +180,10 @@ void StaticLocationWidget::onClick() {
 
 	if (_soundMouseClick) {
 		_soundMouseClick->play();
+		// Ensure the click sound is played completely
+		while (_soundMouseClick->isPlaying()) {
+			g_system->delayMillis(10);
+		};
```
sev-: You need to add g_system->updateScreen() here too, otherwise you could get artefacts or jerky mouse movements.

.gitignore
```diff
@@ -169,3 +169,11 @@ gmon.out
...
 \#*
 TAGS
+
+#Ignore Visual Studio Code files
+/.vscode
+
+#Ignore testing .dll files
+libogg.dll
+libvorbis.dll
+libvorbisfile.dll
```
sev-: No newline at the end of the file

engines/stark/ui/menu/locationscreen.cpp
```diff
@@ -171,6 +180,11 @@ void StaticLocationWidget::onClick() {
 
 	if (_soundMouseClick) {
 		_soundMouseClick->play();
+		// Ensure the click sound is played completely
+		while (_soundMouseClick->isPlaying()) {
+			g_system->delayMillis(10);
+			g_system->updateScreen();
+		};
```
bgK: Please remove that extra semicolon.

DouglasLiuGamer:
A rebasing has been done to sync with the master.

PR #1418, opened 2018-05-23
MYST3: Add Autosave Support

dafioram:
Autosave is enabled and saved to Autosave.M3S
if that file doesn't exist or is an autosave.

Autosave is done on exit or based on the save
interval specified in the config file.

### DIFF
--- a/engines/myst3/myst3.cpp
+++ b/engines/myst3/myst3.cpp
@@ -71,7 +71,7 @@ Myst3Engine::Myst3Engine(OSystem *syst, const Myst3GameDescription *version) :
 		_inputSpacePressed(false), _inputEnterPressed(false),
 		_inputEscapePressed(false), _inputTildePressed(false),
 		_inputEscapePressedNotConsumed(false),
-		_interactive(false),
+		_interactive(false), _lastSaveTime(0),
 		_menuAction(0), _projectorBackground(0),
 		_shakeEffect(0), _rotationEffect(0), _backgroundSoundScriptLastRoomId(0),
 		_transition(0), _frameLimiter(0), _inventoryManualHide(false) {
@@ -210,6 +210,9 @@ Common::Error Myst3Engine::run() {
 		drawFrame();
 	}
 
+	if (!_menu->getThumbnailValid())
+		_menu->saveThumbnail(); // Update thumbnail before saving
+	tryAutoSaving(); //Attempt to autosave before exiting
 	unloadNode();
 
 	_archiveNode->close();
@@ -431,6 +434,7 @@ void Myst3Engine::processInput(bool interactive) {
 	}
 
 	bool shouldInteractWithHoveredElement = false;
+	_menu->setThumbnailValid(false);
 
 	// Process events
 	Common::Event event;
@@ -536,6 +540,11 @@ void Myst3Engine::processInput(bool interactive) {
 
 	if (shouldInteractWithHoveredElement && interactive) {
 		interactWithHoveredElement();
+	}
+
+	if (shouldPerformAutoSave(_lastSaveTime)) {
+		_menu->saveThumbnail(); // Update thumbnail before saving
+		tryAutoSaving();
 	}
 
 	// Open main menu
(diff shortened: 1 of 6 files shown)


engines/myst3/myst3.cpp
```diff
@@ -1487,6 +1491,28 @@ bool Myst3Engine::canLoadGameStateCurrently() {
...
+	}
+
+	_lastSaveTime = _system->getMillis();
+
+	if (!_state->isAutoSaveAllowed()) {
+		return; // Can't autosave ever, try again after the next autosave delay
+	}
+
+	Common::String saveName = "Autosave";
+	Common::String fileName = saveName + ".M3S";
```
bgK: The XBox version of the game has incompatible saves when compared to the PC/Mac version. In ResidualVM they have different filename extensions to set them apart. This function can be used to construct the filename with the proper extension:
https://github.com/residualvm/residualvm/blob/9484bbdd21a8e3c08c8e20ffe00b70b15c475ff3/engines/myst3/state.h#L439

engines/myst3/myst3.cpp
```diff
@@ -1487,6 +1491,28 @@ bool Myst3Engine::canLoadGameStateCurrently() {
...
+
+	// Save the state and the thumbnail
+	Common::OutSaveFile *save = getSaveFileManager()->openForSaving(fileName);
+	_state->setSaveDescription(saveName);
```
bgK: I don't think this is correct.
1. The user saves his game with a custom name
2. Autosave happens
3. Suddenly when the user tries to save, his game is named "autosave"

engines/myst3/state.cpp
```diff
@@ -531,7 +535,30 @@ bool GameState::load(Common::InSaveFile *saveFile) {
...
+	if (!saveFile) { // Autosave.M3S doesn't exist or is corrupt
+		return true;
+	}
+
+	// Get autoSave value from saved file
+	StateData data;
+	Common::Serializer s = Common::Serializer(saveFile.get(), 0);
+	data.syncWithSaveGame(s);
+
+	if (data.autoSave==true) { // Autosave.M3S exists and is an autosave
```
bgK: Please merge this if statement into a `return <boolean expression>`. Testing `if (something == true)` is considered poor style.

engines/myst3/state.h
```diff
@@ -48,12 +49,13 @@ enum ViewType {
 
 class GameState {
 public:
-	GameState(const Common::Platform platform, Database *database);
+	GameState(const Common::Platform platform, Database *database, Myst3Engine *vm);
```
bgK: Game state intentionally did not have a dependency onto `Myst3Engine` for future plans I have with that code. Please  move the part of the code that opens the file to `Myst3Engine`.

engines/myst3/state.cpp
dafioram: Is the change to be:
if (data.autoSave==true) -> if (data.autoSave)?
I think you are also saying:
if (data.autoSave==true) -> return data.autoSave;?

This latter assumes that autosaving should be done if Autosave.M3S exists (even if it is not an autosave) is that okay? If so I may not need to bump the version number or even include an autoSave flag since just checking for the autosave file is sufficient.

bgK: The current code after reading the save is strictly equivalent to `return data.autoSave;`

dafioram:
Done. 

It looks like StateData.saveDescription (a Common::String) is implicitly set to "" on construction? Would it be preferable to set that manually in GameState::StateData::StateData?

bgK:
Thanks for your changes. While testing this, I noticed the autosaves were corrupted and would not load. I think this is because we are trying to save a thumbnail of the screen when none have been captured.

I also noticed nothing was preventing the autosave system from making saves when no game is loaded. `canSaveCurrenlty` needs to be created and to check for that condition (State.LocationRoom == 901 && State.MenuSavedAge==0).

bgK:
> It looks like StateData.saveDescription (a Common::String) is implicitly set to "" on construction? Would it be preferable to set that manually in GameState::StateData::StateData?

It's usual not to explicitly initialize fields non primitive fields when the default value is satisfying.

dafioram:
Weird, when I had lots of saves everything worked fine, but when I started with no saves it has these issues.

dafioram:
Can I change the condition to just !(State.MenuSavedAge==0) (never save when the menu is open)? 

In fact I would like _lastSaveTime to not increment while the menu is pulled up. Otherwise it will autosave as soon as the menu is closed (if the user had the menu opened for a bit).

dafioram:
Updated. 

I believe the first issue of a thumbnail not being saved was solved by fixing the 2nd issue, saving when no game has been started and no saved games.

bgK:
> I believe the first issue of a thumbnail not being saved was solved by fixing the 2nd issue, saving when no game has been started and no saved games.

I'm sorry, but I'm still getting a crash. With no autosave file, start a new game using the main menu. Try to quit ResidualVM:
`residualvm: ../common/ptr.h:152: T* Common::SharedPtr<T>::operator->() const [with T = Graphics::Surface; Common::SharedPtr<T>::PointerType = Graphics::Surface*]: Assertion `_pointer' failed.`

> Can I change the condition to just !(State.MenuSavedAge==0) (never save when the menu is open)?

Mhh, when we end up making autosaves working, there's only one tiny step left to enable saving from the GMM. `canSaveCurrently` is used for that purpose, and I see no reason to forbid saving from the GMM while in the game menu and a game is loaded. Do you agree?

engines/myst3/myst3.cpp
```diff
@@ -1479,6 +1483,10 @@ void Myst3Engine::dragItem(uint16 statusVar, uint16 movie, uint16 frame, uint16
 	}
 }
 
+bool Myst3Engine::canSaveGameStateCurrently() {
+	return canLoadGameStateCurrently() && !(_state->getVar(62) == 901 && _state->getVar(67)==0);
```
bgK: Please use the named versions of the variable accessors for better readability (getLocationRoom and getMenuSavedAge).

dafioram:
Alright let see if you can break this one.

Usually the menu gets called and when that happens the thumbnail is updated, but if you just quit without opening the menu ever then the thumbnail is not created and hence the crash. It took a while to decide were to put the additional logic checks for generating another thumbnail when the menu hasn't been opened (and the game has been loaded). Its possible it could go in Menu::goToNode().

If I just generate a thumbnail it at the end of the program then if the menu had been called it would just save a thumbnail of the menu. Maybe there is some cleaner logic.

I think its fine saving in the menu with gmm, just as long as the thumbnail is generated before. The autosave time should probably freeze while in the menu.

bgK:
Thanks. I merged a modified version of this:
* Rewritten thumbnail handling to better support autosaving
* Removed checking if the autosave file is actually an autosave. As the save system is file name based rather than slot based, this made less sense. Also the original menus make it easy to overwrite the autosave, permanently breaking autosaving.

PR #1421, opened 2018-05-28
MYST3: Turn roomID into an enum

dafioram:
I took the list of room ids in database.cpp and turned them into an enum in myt3.h.

I think the names could be better than what I currently have (from database.cpp). Like kAmateriaStart and kAmateria ... E.x., I don't even know for sure if kMAIS (room id 1002) is Amateria.

It crashes when I start up saying that No room with ID 101! That is in database.cpp, when it searches for all the rooms in the list. It doesn't find kMENU and I don't know why not. 

I am doing lots of static_casts between RoomID and uint16 since the getLocationRoom() and setLocationRoom() are out of my control and I think they are uin16 so some cast(s) is not going correctly.


dafioram:
I should put this on hold until I get it more usable.

PR #1422, opened 2018-05-29
STARK: Implement the save & load menu

DouglasLiuGamer:
This PR is corresponding to implement the save & load menu of *The Longest Journey*.

### DIFF
--- a/engines/stark/services/gamechapter.cpp
+++ b/engines/stark/services/gamechapter.cpp
@@ -1,0 +1,63 @@
+/* ResidualVM - A 3D game interpreter
+ *
+ * ResidualVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the AUTHORS
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "engines/stark/services/gamechapter.h"
+#include "engines/stark/services/services.h"
+#include "engines/stark/services/global.h"
+
+#include "common/archive.h"
+#include "common/stream.h"
+#include "common/tokenizer.h"
+#include "common/debug.h"
+
+namespace Stark {
+
+GameChapter::GameChapter() {
+	Common::SeekableReadStream *stream = SearchMan.createReadStreamForMember("chapters.ini");
+	if (!stream) {
+		error("Opening 'chapters.ini' failed");
+		return;
+	}
+
+	Common::String line, title, subtitle;
+
+	// Assume that the formats of all chapters.ini are the same
+	stream->readLine();
+	while (!stream->eos()) {
+		_chapterEntries.push_back(ChapterEntry());
+
+		line = stream->readLine();
+		Common::StringTokenizer tokens(line, "=:");
+
+		tokens.nextToken();
+		_chapterEntries.back().title = tokens.nextToken();
+		_chapterEntries.back().title.trim();
+		_chapterEntries.back().subtitle = tokens.nextToken();
+		_chapterEntries.back().subtitle.trim();
+	}
+}
+
+int GameChapter::getActualCurrentChapter() {
+	return StarkGlobal->getCurrentChapter() / 10;
+}
+
+} // End of namespace Stark
--- a/engines/stark/services/gamechapter.cpp
+++ b/engines/stark/services/gamechapter.cpp
@@ -24,35 +24,40 @@
 #include "engines/stark/services/services.h"
 #include "engines/stark/services/global.h"
 
-#include "common/archive.h"
-#include "common/stream.h"
 #include "common/tokenizer.h"
-#include "common/debug.h"
+#include "common/ini-file.h"
 
 namespace Stark {
 
 GameChapter::GameChapter() {
-	Common::SeekableReadStream *stream = SearchMan.createReadStreamForMember("chapters.ini");
-	if (!stream) {
-		error("Opening 'chapters.ini' failed");
+	Common::INIFile file;
+	if (!file.loadFromFile("chapters.ini")) {
+		error("Opening file 'chapters.ini' failed");
 		return;
 	}
 
-	Common::String line, title, subtitle;
+	Common::String section = file.getSections().front().name;
+	
+	int index = 0;
+	Common::String key = Common::String::format("%02d", index);
+	Common::String value;
 
-	// Assume that the formats of all chapters.ini are the same
-	stream->readLine();
-	while (!stream->eos()) {
+	while (file.hasKey(key, section)) {
+		file.getKey(key, section, value);
 		_chapterEntries.push_back(ChapterEntry());
 
-		line = stream->readLine();
-		Common::StringTokenizer tokens(line, "=:");
-
-		tokens.nextToken();
+		Common::StringTokenizer tokens(value, ":");
 		_chapterEntries.back().title = tokens.nextToken();
 		_chapterEntries.back().title.trim();
 		_chapterEntries.back().subtitle = tokens.nextToken();
 		_chapterEntries.back().subtitle.trim();
+
+		++index;
+		key = Common::String::format("%02d", index);
+	}
+
+	if (index < 15) {
+		error("File 'chapters.ini' is incomplete");
 	}
 }
 
--- a/engines/stark/stark.cpp
+++ b/engines/stark/stark.cpp
@@ -336,7 +336,8 @@ Common::Error StarkEngine::loadGameState(int slot) {
 
 bool StarkEngine::canSaveGameStateCurrently() {
 	// Disallow saving when there is no level loaded or when a script is running
-	return _global->getLevel() && _userInterface->isInteractive();
+	// or when the save & load menu is currently displayed
+	return _global->getLevel() && _userInterface->isInteractive() && !_userInterface->isInSaveLoadMenuScreen();
 }
 
 Common::Error StarkEngine::saveGameState(int slot, const Common::String &desc) {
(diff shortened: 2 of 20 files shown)


DouglasLiuGamer:
The current save & load menu would render the slot 0 savedata's thumbnail, if there is one. Clicking it shall load it or replace it with a new savedata with preset description.

engines/stark/stark.cpp
```diff
@@ -336,7 +336,8 @@ Common::Error StarkEngine::loadGameState(int slot) {
 
 bool StarkEngine::canSaveGameStateCurrently() {
 	// Disallow saving when there is no level loaded or when a script is running
-	return _global->getLevel() && _userInterface->isInteractive();
+	// or when the save & load menu is currently displayed
+	return _global->getLevel() && _userInterface->isInteractive() && !_userInterface->isInSaveLoadMenuScreen();
 }
```
DouglasLiuGamer: I used to try an implementation that allows saving through GMM while the save & load menu is displayed, but it is kind of complex since the menu needs to be notified when a save slot is updated and update the thumbnail of the corresponding widget. Maybe just disable the saving is a better approach?

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -136,6 +147,8 @@ SaveDataWidget::SaveDataWidget(int slot, Gfx::Driver *gfx, SaveLoadMenuScreen *s
 
 		Graphics::Surface *thumb = metadata.readGameScreenThumbnail(&stream);
 		_texture->update(thumb);
+
+		delete thumb;
 	}
 }
```
DouglasLiuGamer: I feel like putting a `delete save;` here is needed, but it will crash the game. The `StarkEngine::loadGameState()` also doesn't delete the `save`, why is that?

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -113,7 +131,16 @@ void SaveMenuScreen::open() {
...
 
 void SaveMenuScreen::onSlotSelected(int slot) {
-	g_engine->saveGameState(slot, "TestSave");
+	int chapter = StarkGlobal->getCurrentChapter();
+
```
DouglasLiuGamer: What does the returned value of `getCurrentChapter()` really mean? I thought it was the index of the chapter, but calling this function in *Chapter 1* gives me value `10`.

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -132,11 +159,13 @@ SaveDataWidget::SaveDataWidget(int slot, Gfx::Driver *gfx, SaveLoadMenuScreen *s
...
-		_surfaceRenderer(gfx->createSurfaceRenderer()) {
-	// Obtain the thumbnail
+		_surfaceRenderer(gfx->createSurfaceRenderer()),
+		_textDesc(gfx),
+		_textTime(gfx),
+		_isMouseHovered(false) {
+	// Load the corresponding save slot data
 	Common::String filename = StarkEngine::formatSaveName(ConfMan.getActiveDomainName().c_str(), _slot);
 	Common::InSaveFile *save = g_system->getSavefileManager()->openForLoading(filename);
 	if (save) {
```
DouglasLiuGamer: I feel like putting a `delete save;` at the end is needed, but it will crash the game. The `StarkEngine::loadGameState()` also doesn't delete the `save`, why is that?

DouglasLiuGamer:
**Beware**
I change the `SaveMetaData` class to include the second part of the saving time, which may invalidate old save data. New save data may also have problems loading to old ResidualVM.

bgK:
> I change the SaveMetaData class to include the second part of the saving time, which may invalidate old save data. New save data may also have problems loading to old ResidualVM.

Hi, Please increment the save version number and load the seconds for saves that have at least the new version number. This way old saves remain compatible.

https://github.com/residualvm/residualvm/blob/76c628a25f83c470afd074ae0a585352a7ffb8e9/engines/stark/services/stateprovider.h#L127

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -0,0 +1,256 @@
...
+		// Obtain the text
+		_textDesc.setText(metadata.description);
+		_textDesc.setColor(_textColor);
+		_textDesc.setFont(FontProvider::FontType::kCustomFont, 3);
+
+		_textTime.setText(Common::String::format("%02d:%02d:%02d %02d/%02d/%02d",
+				metadata.saveHour, metadata.saveMinute, metadata.saveSecond,
+				metadata.saveMonth, metadata.saveDay, metadata.saveYear % 100));
+		_textTime.setColor(_textColor);
+		_textTime.setFont(FontProvider::FontType::kCustomFont, 3);
```
dafioram: I think you want FontProvider::kCustomFont instead of FontProvider::FontType::kCustomFont.

engines/stark/ui/menu/saveloadmenu.cpp
dafioram: The constructor for StateReadStream() is two arguments, the file and a flag (DisposeAfterUse::YES and DisposeAfterUse::NO) the flag indicates if the input file should be deleted or not after use. 

In your case you used it with 1 argument which looks like that defaults to the 2 argument case with the flag set to dispose. I am not exactly sure where that is happening since the constructor to SubReadStream defaults the dispose to NO (see common/substeam.h). StateReadStream is a SeekableSubReadStream which is also a SubReadStream.

So you don't need to delete save since that is already being done by the destructor of StateReadStream, but just to be clear you may want to explicitly set the 2nd argument to be DisposeAfterUse::YES in the constructor of StateReadStream.

Alternately, you can set it to NO and delete the save manually.

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+void SaveMenuScreen::onSlotSelected(int slot) {
+	int chapter = StarkGlobal->getCurrentChapter();
+
+	Common::String desc;
+	if (chapter == 0) {
+		desc = "Prologue";
+	} else {
+		desc = Common::String::format("Chapter %d", chapter);
```
bgK: Ultimately, chapter names will have to come from `chapters.ini` so they are localized. But I guess having them hardcoded is fine for the first version. The original engine uses the string before the colon in `chapters.ini`.

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -0,0 +1,276 @@
...
+		StateReadStream stream(save);
+		Common::ErrorCode metadataErrorCode = metadata.read(&stream, filename);
+		if (metadataErrorCode != Common::kNoError) {
+			error("Unable to read save metadata with error code %d.", metadataErrorCode);
+		}
+
+		// Obtain the thumbnail
+		Graphics::Surface *thumb = metadata.readGameScreenThumbnail(&stream);
+		_texture->update(thumb);
+		delete thumb;
```
bgK: `Graphics::Surface` is very stupid. `->free()` needs to be called as well.

DouglasLiuGamer:
@bgK Currently `SaveDataWidget` will load and update its thumbnail and text only when it is created and clicked on the save menu. Obviously it is unrealistic to update the content on every frame. But the GMM provides saving and deleting operations, which will invalidate the current menu screen's content. Implementing the updating on the change from GMM is going to be a complex task, and it feels like unnecessary for such a strange behaviour. ~I mean, who on earth will open two save & load menus just for fun?~

So, is it okay to just block both saving and loading from GMM when the game is in save & load menu?

p.s. Currently, the saving is already blocked. Deleting a save data through GMM while the load menu is displayed will make the corresponding widget, if it's on the current page, unresponsive.

DouglasLiuGamer:
A new service for retrieving the chapter title and subtitle from `chapters.ini` is created. This would also be helpful when implementing the conversation log, I guess.

bgK:
> So, is it okay to just block both saving and loading from GMM when the game is in save & load menu?

Sure, ok.

engines/stark/services/gamechapter.cpp
```diff
@@ -0,0 +1,63 @@
...
+GameChapter::GameChapter() {
+	Common::SeekableReadStream *stream = SearchMan.createReadStreamForMember("chapters.ini");
+	if (!stream) {
+		error("Opening 'chapters.ini' failed");
+		return;
+	}
+
+	Common::String line, title, subtitle;
+
+	// Assume that the formats of all chapters.ini are the same
```
bgK: Did you notice `Common::INIFile`?

DouglasLiuGamer: Oops, didn't see that. I'll take a look.

DouglasLiuGamer:
The current mechanism of capturing the screenshot will also capture the cursor. I have spent some time trying to get rid of it, but the best I could get will also remove the model from the thumbnail. Any suggestion?

bgK:
> The current mechanism of capturing the screenshot will also capture the cursor. I have spent some time trying to get rid of it, but the best I could get will also remove the model from the thumbnail. Any suggestion?

A simple way would be to re-render the game window only just before capturing the screenshot. (And maybe re-render the whole screen afterwards if that causes glitches)

DouglasLiuGamer:
> A simple way would be to re-render the game window only just before capturing the screenshot.

Already tried this before. Simply re-render the game window (or even the game screen) will cause only the background rendered, without the models. Calling `updateScreen()` after will also render the cursor.

Still can't figure out what causes this.

bgK:
> Already tried this before. Simply re-render the game window (or even the game screen) will cause only the background rendered, without the models. 

Oh, right. The depth buffer needs to be cleared for the depths tests to work correctly. Calling `StarkGfx->clearScreen` should do it.

engines/stark/ui/menu/diaryindex.cpp
```diff
@@ -94,7 +94,7 @@ void DiaryIndexScreen::open() {
 	_widgets.push_back(new StaticLocationWidget(
 			"Back",
 			CLICK_HANDLER(DiaryIndexScreen, backHandler),
-			MOVE_HANDLER(DiaryIndexScreen, widgetTextColorHandler)));
+			nullptr));
```
DouglasLiuGamer: Just found out that this text will not change to blue in the original game.

engines/stark/ui/menu/saveloadmenu.cpp
```diff
@@ -265,10 +265,12 @@ void SaveDataWidget::loadSaveDataElements() {
...
-		Graphics::Surface *thumb = metadata.readGameScreenThumbnail(&stream);
-		_texture->update(thumb);
-		thumb->free();
-		delete thumb;
+		if (metadata.version >= 9) {
+			Graphics::Surface *thumb = metadata.readGameScreenThumbnail(&stream);
+			_texture->update(thumb);
+			thumb->free();
+			delete thumb;
+		}
```
DouglasLiuGamer: I don't have save data older than 9. Could you please check it for me?

bgK: It's working, thanks!

This is my collection of old saved games in case you need it: https://filenurse.com/download/361b5da097c32696eff053a7b026ad7b.html

A lot of them were made back when some entity save handlers were missing, so they don't load all the state, but oh well...

Issue #1423, opened 2018-05-29
STARK: Credits doesn't work without High Quality Video (GOG version)


Faalagorn:
Steam version and Steam demo doesn't have option to uncheck the high quality videos, plus Steam demo doesn't have credits at all, so I can't check that, but for the full GOG version unchecking the *High Quality Video* option makes the credits unable to play – not sure if it's due to lack of high quality video of credits in game file and if it is intentional or not, but I thought I'll let it now in case it's an overlook.

DouglasLiuGamer:
Strange. I am also using the Steam version for development, which does not have any low-resolution videos. In this case, the settings menu should not have the video option, so normally you can't *uncheck* it unless you manually change it in the configuration file.

Could you please take a look at your Steam version's game data, go to the **Global/xarc** folder, and see if there is any **.bbb** file with the suffix "_lo_res" in its name? If it does, does it have **1402_lo_res.bbb**? I use this FunCom logo video to check whether there are low-resolution videos.

Faalagorn:
Oops, my bad – i confused it – it was *GOG* version that had the videos, while Steam indeed doesn't have the option present as well, so the bug is for the GOG version, not the *Steam* version :)

Faalagorn:
I found out that ***0e02.bbb** file is the one with credits and there's indeed no *lo_res* version of it distributed with GOG files. Steam full version and Steam demo doesn't have low res files at all, so it seems to be game's version issue. I'm not sure if any of the other versions have a low resolution credits, and how the original game behaves in that case.

DouglasLiuGamer:
It will be unrealistic to check all the videos to determine whether to provide the setting... Anyway, I made a PR which will use the original video when low-res one cannot be found. There is a cost added for such an additional checking, but if that's the case for all GOG versions, I think we have to cover this.

Faalagorn:
Not sure how computational heavy is this, but This should also help if for some reason someone manually change it in configuration files like you said before – though it would be worth checking how the original game behaves with that, even if just for curiosity.

DouglasLiuGamer:
Now that I think of it there is almost no computational heaviness actually... I thought too much.

My original implementation can well handle the case of changing the configuration files manually. The GOG version you mentioned is just kind of strange, whose videos are not consistent with only the credits video lacking low-res version. Anyway, if this exists, we need to handle it.

Faalagorn:
Confirmed working as of 388eafa (tested with GOG version).

PR #1425, opened 2018-05-30
STARK: Use high-res video when low-res is not available

DouglasLiuGamer:
Related to https://github.com/residualvm/residualvm/issues/1423

When the *High Quality Videos* setting is unchecked but the low-resolution video cannot be found, the original high-resolution video, if it exists, will be played instead.

This is done in case some versions of TLJ have inconsistent videos.

### DIFF
--- a/engines/stark/ui/world/fmvscreen.cpp
+++ b/engines/stark/ui/world/fmvscreen.cpp
@@ -54,13 +54,30 @@ FMVScreen::~FMVScreen() {
 	delete _surfaceRenderer;
 }
 
-void FMVScreen::play(Common::String name) {
+void FMVScreen::play(const Common::String &name) {
+	Common::SeekableReadStream *stream;
+
+	// Play the low-resolution video, if possible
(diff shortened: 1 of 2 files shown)


engines/stark/ui/world/fmvscreen.cpp
```diff
@@ -54,13 +54,30 @@ FMVScreen::~FMVScreen() {
...
 	if (!StarkSettings->getBoolSetting(Settings::kHighFMV) && StarkSettings->hasLowResFMV()) {
-		name.erase(name.size() - 4);
-		name += "_lo_res.bbb";
+		Common::String lowResName = name;
+		lowResName.erase(lowResName.size() - 4);
+		lowResName += "_lo_res.bbb";
+
+		stream = StarkArchiveLoader->getExternalFile(lowResName, "Global/");
+		if (stream) {
+			_decoder->loadStream(stream);
```
bgK: Please improve the code organization so this section is not duplicated with the high-resolution variant.

bgK:
Thanks!

PR #1428, opened 2018-05-31
MYST3: Add Saving from GMM

dafioram:
The save dialogue can be opened by F5.
The in game menu can still be opened
by pressing escape.

The Menu object internally manage a thumbnail
for saved games. But now a function has been
added to Menu to have a new thumbnail generated.

A saveGame function has been added to Myst3
that just grabs this already generated thumbnail.

### DIFF
--- a/engines/myst3/myst3.cpp
+++ b/engines/myst3/myst3.cpp
@@ -31,6 +31,7 @@
 
 #include "gui/debugger.h"
 #include "gui/error.h"
+#include "gui/saveload.h"
 
 #include "engines/engine.h"
 
@@ -139,6 +140,7 @@ bool Myst3Engine::hasFeature(EngineFeature f) const {
 	return
 		(f == kSupportsRTL) ||
 		(f == kSupportsLoadingDuringRuntime) ||
+		(f == kSupportsSavingDuringRuntime) ||
 		(f == kSupportsArbitraryResolutions && !softRenderer);
 }
 
@@ -492,6 +494,7 @@ void Myst3Engine::processInput(bool interactive) {
 				break;
 			case Common::KEYCODE_F5:
 				// Open main menu
+
 				if (_cursor->isVisible() && interactive) {
 					if (_state->getLocationRoom() != 901)
 						_menu->goToNode(100);
@@ -502,7 +505,7 @@ void Myst3Engine::processInput(bool interactive) {
 					_console->attach();
 					_console->onFrame();
 				}
-				break;
+				break;				
 			default:
 				break;
 			}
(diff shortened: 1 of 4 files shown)


engines/myst3/myst3.cpp
```diff
@@ -1505,16 +1507,20 @@ void Myst3Engine::tryAutoSaving() {
...
 
 	// Open the save file
 	Common::String saveName = "Autosave";
+	saveGame(saveName);
+
+}
+
+void Myst3Engine::saveGame(Common::String saveName) {
+	
+	Common::DisposablePtr<Graphics::Surface, Graphics::SurfaceDeleter> thumbnail(nullptr, DisposeAfterUse::NO);
```
bgK: There is no need to use `DisposablePtr` here, since ownership remains to the menu (we're just borrowing a read only reference). `const Graphics::Surface *` should be used.

engines/myst3/myst3.cpp
```diff
@@ -1505,16 +1507,20 @@ void Myst3Engine::tryAutoSaving() {
...
 	// Get a thumbnail of the game screen
-	Common::DisposablePtr<Graphics::Surface, Graphics::SurfaceDeleter> thumbnail(nullptr, DisposeAfterUse::NO);
-	if (_menu->isOpen()) {
-		thumbnail.reset(_menu->borrowSaveThumbnail(), DisposeAfterUse::NO);
-	} else {
-		// Capture thumbnail before saving
-		thumbnail.reset(_menu->captureThumbnail(), DisposeAfterUse::YES);
-	}
+	if (!_menu->isOpen())
+		_menu->generateSaveThumbnail();
```
bgK: The previous code was careful not to leave the thumbnail memory allocated.

engines/myst3/myst3.cpp
```diff
@@ -1877,6 +1900,17 @@ void Myst3Engine::pauseEngineIntern(bool pause) {
 
 	_state->pauseEngine(pause);
 
+	// Grab a game screen thumbnail in case we need one when writing a save file
+	// Get a thumbnail of the game screen
+	
+
+	if (_state->getLocationRoom() != 901) {
+		if (pause) {
```
bgK: You should also be able to free the thumbnail memory if the game menu is not open when leaving the GMM.

dafioram: I am having the Menu object manage the thumbnails memory. 

If a new thumbnail is generated (generateSaveThumbnail() is called) then a reset(new thumbnail) is done in menu which deletes the old and replaces the new one. When the program exits that scoped pointer deletes the thumbnail if it is present in the Menu object. This is what I have intended to do at least.

I would prefer to not have to manage the logic for deletion of the thumbnail outside of the Menu class.

Is that sufficient?

dafioram:
Updated. Save dialog opened with ctrl+s. Empty save names default to "Saved Game", like the original does. So if you save using the gmm twice in an empty slot with no description it will save to "Saved Game" Twice.

I cannot save the file description using slc.createDefaultSaveDescription(slot) because that causes a crash, something about a "\".

I call _menu->clearSaveThumbnail() in Myst3::saveGame() which clears the thumbnail for saving once I have used it to create a saved game. If I clear it in pauseEngineIntern then it will not be available to use in Myst3::saveGame().

dafioram:
I've removed the clearSaveThumbnail function since that causes a segfault when it tries to save a saved game when there is no thumbnail present. I now a check to see if the thumbnail is valid before saving. 

If you try to save over another saved game using crtrl+s it will save it as saved game.m3s.m3s so I need to remove the extension when overriding a save.

dafioram:
For saving via ctrl+s I've addressed the case when the user triess to save over an already existing save or if the user tries to save a save name with a .M3S or .M3X. Those cases now have the extension striped from the name and then added back on. 

Previously, if save.m3s was saved and the user tried to override it then it would make a new save save.m3s.m3s.

engines/myst3/myst3.cpp
```diff
@@ -1505,28 +1512,43 @@ void Myst3Engine::tryAutoSaving() {
...
-		warning("Unable to open the autosave file: %s.", fileName.c_str());
-	}
+	saveGame(saveName);
 
-	// Save the state and the thumbnail
-	if (!_state->save(save, saveName, thumbnail.get())) {
-		warning("Unable to write the autosave to '%s'.", fileName.c_str());
+}
+
+void Myst3Engine::saveGame(Common::String &saveName) {
```
bgK: Is there any reason why you're not implementing `Common::Error saveGameState(int slot, const Common::String &desc)` and adding `kSupportsSavingDuringRuntime` to the features list so saving actually works from the GMM?

dafioram: The game crashes if I try to save without a name (using the auto generated descriptor based on timestamps). 

Should I write my own MainMenuDialog::save()?

residualvm: backends/fs/posix/posix-fs.cpp:113: virtual AbstractFSNode* POSIXFilesystemNode::getChild(const Common::String&) const: Assertion `!n.contains('/')' failed.
Aborted (core dumped)

bgK: I think it crashes because the filename has the following form when using the default save description: `%04d-%02d-%02d / %02d:%02d:%02d`.  The slash is not allowed in filenames.

IMO the engine should replace invalid characters with something harmless in `saveGameState` or disallow save names that contain invalid characters like the Pegasus engine in ScummVM: https://github.com/scummvm/scummvm/blob/4007a7088a1c4266abfe466540a70711b9050926/engines/pegasus/pegasus.cpp#L715

dafioram:
That did the trick.

This revision should be pretty solid.

bgK:
Thanks for your contribution. I merged this manually, making some changes:
- Improved error handing
- I reused the same save filename check code as the Pegasus engine in ScummVM, as there are other potentially forbidden characters across platforms. Especially on Windows.
- I removed the default save name for the original menu, as I think it does not look good on the localized versions of the game.

dafioram:
Thanks.

PR #1429, opened 2018-06-03
STARK: Pause the game's sound when the game screen is closed

DouglasLiuGamer:
In the original game, all the sounds of the game screen will be paused when the menu is opened. Simply pausing through the `mixer` seems to be enough and the sounds of the menu are unaffected.

### DIFF
--- a/engines/stark/ui/world/gamescreen.cpp
+++ b/engines/stark/ui/world/gamescreen.cpp
@@ -32,6 +32,9 @@
 #include "engines/stark/ui/world/inventorywindow.h"
 #include "engines/stark/ui/world/topmenu.h"
 
+#include "engines/engine.h"
+#include "audio/mixer.h"
+
 namespace Stark {
 
 GameScreen::GameScreen(Gfx::Driver *gfx, Cursor *cursor) :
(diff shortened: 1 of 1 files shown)


bgK:
Sounds are not the only thing that need to be paused when entering the menu. Videos need to be paused as well (otherwise they try to catch up for the time when the menu was open when leaving the menu. Probably other things as well (timed events, ...). I'd prefer a more general way of pausing the game world.

Maybe the `onEnginePause` entity event can be reused, adding a handler to the `Sound` entity to pause the corresponding mixer channel.

See also: https://github.com/residualvm/residualvm/blob/a27834fa7644570dc5898ad88732b42bf0ef0b6c/engines/stark/stark.cpp#L391

engines/stark/ui/world/gamescreen.cpp
```diff
@@ -65,14 +65,15 @@ GameScreen::~GameScreen() {
...
+	}
 	StarkUserInterface->freeGameScreenThumbnail();
 }
 
 void GameScreen::close() {
 	_cursor->setMouseHint("");
-	g_engine->_mixer->pauseAll(true);
-	StarkUserInterface->saveGameScreenThumbnail();
+	g_engine->pauseEngine(true);
 }
```
DouglasLiuGamer: It seems that the original `pauseEngine()` is well enough. No need to called `saveGameScreenThumbnail()` now since it will be called in the `pauseEngine()`.

DouglasLiuGamer:
Don't merge this now. Have some problem with the thumbnail in another branch. Will cause conflict.

DouglasLiuGamer:
Change on this PR has been merged to [another PR](https://github.com/residualvm/residualvm/pull/1442).

PR #1433, opened 2018-06-03
DOCS: Updated Debugging section in README

Faalagorn:
Added description of commands common to the engines, moved `openlog` command there and moved and expanded the information how to use debugging console to the general section, as it's the same for all engines. Also indicated that for TLJ all commands are listed while for other games only the most useful are.


bgK:
Thanks! I have merged this manually, slightly altering the wording to make it clear the commands are debug console commands.

PR #1434, opened 2018-06-03
STARK: Updated descriptions of some debug commands

Faalagorn:
* Removed full stops from the last setence in `debugPrintf`s to unifty the style
* `dump` folder for `dumpArchive` must exist in location ResidualVM is launched from (e.g. $HOMEDIR)
* `dumpArchive` require not only name of the archive but (relative) path to it, e.g. `4f/04/04.xarc` instead of simply `04.xarc`
* Indicated to use `listInventory`  to get ids for `enableInventoryItem`
* `enableInventoryItem` incorrectly had description from `changeLocation`
* Indicated to use `listLocations`  to get ids for `changeLocation`

### DIFF
--- a/engines/stark/console.cpp
+++ b/engines/stark/console.cpp
@@ -72,10 +72,10 @@ Console::~Console() {
 
 bool Console::Cmd_DumpArchive(int argc, const char **argv) {
 	if (argc != 2) {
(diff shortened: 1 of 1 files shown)


bgK:
Thanks!

PR #1449, opened 2018-06-15
MYST3: Minor cursor refactoring

dafioram:
Added some comments and made a loop use the size
rather than hardcoded number.


bgK:
Thanks!

PR #1450, opened 2018-06-16
STARK: Implement the conversation log menu

DouglasLiuGamer:
This PR is corresponding to implementing the conversation log menu in TLJ.


DouglasLiuGamer:
@bgK The whole implementation is basically finished and it's ready for proofreading.

I find it extremely difficult to make the layout of widgets **exactly** the same as the original, at least not in a reasonable way. I suspect that this is due to the difference of the graphics layers. In https://github.com/residualvm/residualvm/pull/1450/commits/93327fa1fe7714995de9c27c055bbdfba1542e9e I did my best to make the whole layout very similar. Differences should be unnoticeable unless you put it together with the original. If you think it is not satisfactory then I think we'll need to take a deep look at the disassembly.

engines/stark/ui/menu/dialogmenu.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+DialogLineText::DialogLineText(Gfx::Driver *gfx, uint logIndex, uint lineIndex, uint boxWidth) :
+		_namePos(), _linePos(), _boxWidth(boxWidth),
+		_nameText(gfx), _lineText(gfx) {
+	Diary::ConversationLogLine logLine = StarkDiary->getDialog(logIndex).lines[lineIndex];
+
+	Common::String name = StarkGlobal->getCharacterName(logLine.characterId);
+	name.toUppercase();
+
+	uint color = name == "APRIL" ? _textColorApril : _textColorNormal;
```
bgK: This is how this test is done in other parts of the engine: https://github.com/residualvm/residualvm/blob/8b5e7433f5cf1a17205c1caa9b5f3788f405e45b/engines/stark/resources/speech.cpp#L131

engines/stark/ui/menu/dialogmenu.cpp
```diff
@@ -0,0 +1,386 @@
...
+		height += dialogTitleWidget->getHeight();
+
+		if (dialogTitleWidget->getChapter() != _curMaxChapter) {
+			_curMaxChapter = dialogTitleWidget->getChapter();
+			chapterTitleText = new ChapterTitleText(_gfx, _curMaxChapter);
+			height += chapterTitleText->getHeight() + 2 * space;
+		}
+
+		if (pos.y + height > bottom) {
+			break;
```
bgK: When we break here the objects created above are never freed.

engines/stark/ui/menu/dialogmenu.cpp
```diff
@@ -0,0 +1,386 @@
...
+	Diary::ConversationLog dialog = StarkDiary->getDialog(_curLogIndex);
+
+	while (_nextLineIndex < dialog.lines.size()) {
+		height = 0;
+
+		dialogLineText = new DialogLineText(_gfx, _curLogIndex, _nextLineIndex, boxWidth);
+		height = dialogLineText->getHeight();
+
+		if (pos.y + height + space > bottom) {
+			break;
```
bgK: When we break here `DialogLineText` is not freed.

bgK:
> Differences should be unnoticeable unless you put it together with the original.

It looks close enough to me. I don't think anybody will mind if it's a few pixels of the original in this case.

DouglasLiuGamer:
@bgK If you also think it's okay, we can merge.

Issue #1435, opened 2018-06-03
STARK: Segmentation fault or crash when loading wrong level/location with changeLocation


Faalagorn:
The Longest Journey crashes with `Segmentation fault (core dumped)` error printed to stdout when trying to load a non-existing level (e.g. `changeLocation ff ff`) or with  `Unexpected resource type when casting resource Level instead of Location!` printed to stdout when level is right but location is wrong (e.g. `changeLocation 0a ff`.

I think additional check should be performed when these commands are initiated to check whether the level or location exist and if not, just print an error to console instead stdout and do not crash ResidualVM.

* Information about the game: The Longest Journey, all versions (tested on GOG, Steam, Steam demo, retail demo)
* Language of game: English
* Platform: Linux

DouglasLiuGamer:
Fixed in #1456

Issue #1432, opened 2018-06-03
STARK: enableInventoryItem crashes the game when used with non-existing id


Faalagorn:
When you type `enableInventoryItem` with an id that does not exist, ex. `enableInventoryItem 107` or `enableInventoryItem -1` into the game debug console (`ctrl + d`), the game crashes with the following error message:
```
residualvm: ./common/array.h:192: T& Common::Array<T>::operator[](Common::Array<T>::size_type) [with T = Stark::Resources::Item*; Common::Array<T>::size_type = unsigned int]: Assertion `idx < _size' failed.
Aborted (core dumped)
```

* Information about the game: The Longest Journey, all versions (tested on GOG, Steam, Steam demo, retail demo)
* Language of game: English
* Platform: Linux

DouglasLiuGamer:
Fixed in #1456

Issue #1431, opened 2018-06-03
STARK: listInventory should print list in debug console instead stdout


Faalagorn:
`listInventory` command from the debug (`ctrl + d`) menu currently print the list of items as `WARNING: Item…` into stdout. It should probably print it's info to the residual's console, just like the `listLocations` and `listScripts` commands do.

* Information about the game: The Longest Journey, all versions (tested on GOG, Steam, Steam demo, retail demo)
* Language of game: English
* Platform: Linux

DouglasLiuGamer:
Fixed in #1456

Issue #1430, opened 2018-06-03
STARK: Segmentation fault when most commands are being used when level is not loaded


Faalagorn:
When you type any of these commands to the debug console (`Ctrl +d`), while the game is at the initial Funcom video, credits, main or setting menu (and probably other menus that are yet to be implemented in `master`), it crashes with the `Segmentation fault (core dumped)` error printed to stdout:
* `changeChapter` + param
* `changeKnowledge` + params
* `changeLocation` + params
* `chapter`
* `decompileScript` + param
* `dumpGlobal`
* `dumpKnowledge`
* `dumpLevel`
*  `dumpLocation`
* `dumpRoot`
* `enableInventoryItem` + param
* `enableScript` + params
* `forceScript` + param
* `listInventory`
* `listScripts`
*  `location`

Commands that don't cause the crash:
* `dumpArchive` + param
* `dumpStatic`
* `listLocations`
* `testDecompiler`

as well as the commands that require parameters when passed without any parameters + built-in commands (`help`, `quit`, `exit`, `openlog`, `debugflag_list`, `debugflag_enable`, `debugflag_disable`, `md5` and `md5mac`).

I think additional check should be performed when these commands are initiated to check whether the level is loaded.

* Information about the game: The Longest Journey, all versions (tested on GOG, Steam, Steam demo, retail demo)
* Language of game: English
* Platform: Linux

Faalagorn:
The same happens on the newly implemented save/load menu, but only when accessed from the main menu, so nothing really changes.

DouglasLiuGamer:
Fixed in #1456

PR #1456, opened 2018-06-24
STARK: Improve the debug console

DouglasLiuGamer:
Issue #1430 #1431 #1432 #1435 

The previous PR is merged by accident. This is a new and clean one.

### DIFF
--- a/engines/stark/console.cpp
+++ b/engines/stark/console.cpp
@@ -477,13 +477,10 @@ bool Console::Cmd_ListInventory(int argc, const char **argv) {
 		return true;
 	}
 
-	bool printAll = argc != 2;
 	Common::Array<Resources::Item*> inventoryItems = inventory->listChildren<Resources::Item>(Resources::Item::kItemInventory);
 	Common::Array<Resources::Item*>::iterator it = inventoryItems.begin();
 	for (int i = 0; it != inventoryItems.end(); ++it, i++) {
-		if (printAll || (*it)->isEnabled()) {
-			debugPrintf("Item %d: %s\n", i, (*it)->getName().c_str());
-		}
+		debugPrintf("Item %d: %s%s\n", i, (*it)->getName().c_str(), (*it)->isEnabled() ? " (enabled)" : "");
 	}
 
 	return true;
@@ -577,7 +574,7 @@ bool Console::Cmd_ChangeLocation(int argc, const char **argv) {
 	// Assert indices
 	Common::String xarcFileName = Common::String::format("%s/%s/%s.xarc", argv[1], argv[2], argv[2]);
 	if (!Common::File::exists(xarcFileName)) {
-		debugPrintf("Invalid location indices %s %s\n", argv[1], argv[2]);
+		debugPrintf("Invalid location %s %s. Use listLocations to get correct indices\n", argv[1], argv[2]);
 		return true;
 	}
 
--- a/engines/stark/console.cpp
+++ b/engines/stark/console.cpp
@@ -116,32 +116,55 @@ bool Console::Cmd_DumpArchive(int argc, const char **argv) {
 }
 
 bool Console::Cmd_DumpRoot(int argc, const char **argv) {
-	StarkGlobal->getRoot()->print();
+	Resources::Root *root = StarkGlobal->getRoot();
+	if (root) {
+		root->print();
+	} else {
+		debugPrintf("The global root has not been loaded\n");
+	}
 
 	return true;
 }
 
 bool Console::Cmd_DumpGlobal(int argc, const char **argv) {
-	StarkGlobal->getLevel()->print();
+	Resources::Level *level = StarkGlobal->getLevel();
+	if (level) {
+		level->print();
+	} else {
+		debugPrintf("The global level has not been loaded\n");
+	}
 
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


engines/stark/console.cpp
```diff
@@ -481,12 +557,57 @@ bool Console::Cmd_ChangeLocation(int argc, const char **argv) {
...
 	uint locationIndex = strtol(argv[2] , nullptr, 16);
 
-	StarkResourceProvider->requestLocationChange(levelIndex, locationIndex);
+	bool isIndicesValid = true;
+
+	// Temporarily replace the global archive loader
+	ArchiveLoader *archiveLoader = new ArchiveLoader();
+	ArchiveLoader *gameArchiveLoader = StarkArchiveLoader;
+	StarkArchiveLoader = archiveLoader;
+
```
DouglasLiuGamer: I spent some time deciding how to asset these two indices. One way is to assert them in the actual `requestLocationChange()`, but the codes there seems to be hard to revert when indices are invalid. So I decided to add an additional load and check here. This will load the same resources twice if indices are correct, but codes are easier to read and since the debug console is an extra feature, it should not be an overhead.

DouglasLiuGamer:
The command `ListLocations` will cause many warning in stdout. I think it is because the debug console doesn't have enough buffer for all the texts. I haven't figure out where to change the buffer size and it doesn't seem to be a good idea to modify the `GUI::Debugger` class since it is used by other games. Any idea?

engines/stark/services/global.cpp
```diff
@@ -52,18 +55,24 @@ void Global::setCurrentChapter(int32 value) {
 }
 
 void Global::printInventory(bool printAll) {
+	// Assume that this function is only called by the debug console
 	Common::Array<Resources::Item*> inventoryItems = _inventory->listChildren<Resources::Item>(Resources::Item::kItemInventory);
 	Common::Array<Resources::Item*>::iterator it = inventoryItems.begin();
 	for (int i = 0; it != inventoryItems.end(); ++it, i++) {
 		if (printAll || (*it)->isEnabled()) {
-			warning("Item %d: %s", i, (*it)->getName().c_str());
+			g_engine->getDebugger()->debugPrintf("Item %d: %s\n", i, (*it)->getName().c_str());
```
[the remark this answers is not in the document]
bgK: I think it would be better to move those two functions to the Console class. It looks weird using `g_engine->getDebugger()` here.

engines/stark/console.cpp
bgK: This seems too complicated to me for the value it brings. IMO, there should be something like `StarkArchiveLoader->canLoadArchive(levelIndex, locationIndex)` that simply checks if the location file exists.

engines/stark/console.cpp
```diff
@@ -554,52 +554,19 @@ bool Console::Cmd_ChangeLocation(int argc, const char **argv) {
...
-	// Assert the location index
-	Resources::Location *location = level->findChildWithIndex<Resources::Location>(locationIndex);
-	if (!location) {
-		debugPrintf("Invalid location index %d, only %d indices available\n", locationIndex,
-				level->listChildren<Resources::Location>().size());
-		isIndicesValid = false;
+	// Assert indices
+	Common::String xarcFileName = Common::String::format("%s/%s/%s.xarc", argv[1], argv[2], argv[2]);
+	if (!Common::File::exists(xarcFileName)) {
+		debugPrintf("Invalid location indices %s %s\n", argv[1], argv[2]);
```
DouglasLiuGamer: The code is a lot simpler, but now we cannot display the range when indices are invalid. Will it be a problem?

engines/stark/console.cpp
```diff
@@ -468,17 +470,29 @@ bool Console::Cmd_DumpLocation(int argc, const char **argv) {
...
-		StarkGlobal->printInventory(argc != 2);
-	} else {
+	Resources::KnowledgeSet *inventory = StarkGlobal->getInventory();
+
+	if (!inventory) {
 		debugPrintf("The inventory has not been loaded\n");
+		return true;
+	}
+
+	bool printAll = argc != 2;
```
DouglasLiuGamer: I kept this parameter as before but I am really confused about its meaning. The whole "adding an argument after `listInventory` will print out enabled inventory items" seems kind of weird. Can we change it to having a `-all` argument or a make a new `listAllInventory` command?

DouglasLiuGamer:
Some commands like `listLocations` and `listInventory` will generate many warnings in stdout which a text "Widget <> has clipping area y < 0". At first I thought it is due to not enough buffer but I'm not sure now. Any idea?

Also the current buffer is not enough for the `listLocations` command. Is it okay to enlarge the buffer of the debug console? Not sure about this since other games also used the same debugger base class.

engines/stark/console.cpp
bgK: The error message could invite the user to use the `listLocations` command to obtain a list of valid locations.

engines/stark/console.cpp
bgK: Sure, feel free to improve this command. Another approach would be to always list all the inventory items and append `(possessed)` or something to the enabled inventory items.

DouglasLiuGamer:
@bgK If there's nothing left I think we can merge them.

PR #1458, opened 2018-06-27
MYST3: Add RoomID enum to help working with Rooms

dafioram:
No behavior changes.

Not all rooms were given useful names, but the most used
were. More useful names for the more obscure rooms can be added
later.


bgK:
Thanks! I have merged this manually, moving stuff around a bit to reduce header dependencies.

PR #1459, opened 2018-06-27
STARK: Updated console message to unify names

Faalagorn:
* Changed "ids" to "indices" to unify the names in console messages about listLocations


bgK:
Thanks

PR #1460, opened 2018-06-27
[STARK] Changed listInventory → listInventoryItems 

Faalagorn:
Changed listInventory to listInventoryItems to fit other commands, that is.:
* changeLocation → listLocations
* decompileScript, enableScript, forceScript → listScripts
so
* enableInventoryItem → listInventory**Items**

Also updated README.md with the updated command name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,7 +420,7 @@ Command               | Description
 `enableInventoryItem` | Enable a specific inventory item
 `enableScript`        | Enable or disable script
 `forceScript`         | Force the execution of a script
-`listInventory`       | List all inventory items in the game
+`listInventoryItems`  | List all inventory items in the game
 `listLocations`       | List all the locations in the game
 `listScripts`         | List all the scripts in current level
 `location`            | Display the current location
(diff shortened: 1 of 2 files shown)


bgK:
Thanks!

(but please follow our commit message formatting conventions)

Faalagorn:
Thanks! You meant the second commit about the README or something else? If that was the second, I was figuring out how to edit two files with the GitHub web interface, and it turned out to not be so straightforward, so that's why I left it at the default without knowing how it'd behave.

Faalagorn:
Okay, I just realized that I used parentheses instead of comma, so I guess that was the reason – I contributed to another project that used similar notation and I confused them, sorry!

Either way, since that was merged already and I realized my previous fix broke the compilation I submitted another PR #1462 – I wasn't sure if I could update this PR instead since it's merged, If so then sorry, I still figure out the intricacies of Git :(.
I should have tested the changes locally before the merge, but I couldn't figure out how to do that from the locally cloned repo I modified (guess I should have forked it locally first), so for this moment I used the git web interface before I figure the proper way to do so from the terminal.

Sorry again for the trouble!

PR #1461, opened 2018-06-28
MYST3: Debug warn instead of error for unknown var

dafioram:
When using the var command in the debug console warn
if the variable number is not in range rather than
error.


bgK:
Thanks!

Issue #1439, opened 2018-06-04
STARK: version info here instead of actual version is shown in the main menu


Faalagorn:
Judging by the comment at #1415 I think it is a known issue, but since that PR is now merged and there's no related issue, I thought I'll make one.

When starting any game (tested with Steam, steam demo, GOG and retail demo), the `version info here`  string is shown instead of actual version info (copyright sign, year, company, build, …).

See more [this comment](1415#issuecomment-389440278) for more info.

* Information about the game: The Longest Journey, all versions (tested on GOG, Steam, Steam demo, retail demo)
* Language of game: English
* Platform: Linux

DouglasLiuGamer:
Yeah, I am aware of this. I have added this one to the project here: https://github.com/residualvm/residualvm/projects/1#card-9934456

Faalagorn:
I saw that TODO note, but wasn't aware what that *info and message box* was actually that thing :smile: . I noticed that so far you're only adding the actual pull requests as references, but if you don't mind, you can add this issue as a reference there as well, so other people looking at it will get an idea what it is as well.

DouglasLiuGamer:
Implemented in #1467

Faalagorn:
I noticed the cursor changes when you hover it over the "©1999 Funcom" string, as it'd be a clickable element. Just noticed it now, not sure if it behaved like that in the older versions and whether or not the cursor changes in the original game?

Too bad the actual version info can not be displayed though. Is it a good idea to hardcode the build versions as well as the string? (for example based on checksum check on EXE or some data files), or even display a residualVM version there?

DouglasLiuGamer:
@Faalagorn 

> the cursor changes when you hover it over the "©1999 Funcom" string

Yes, the original game also behaves like this. I made efforts in ensuring that. Actually, in many places of menus, the cursor will change on texts that are not clickable. That makes it easier to implement the changing mechanism anyway. The only exception is the video replay menu, where the cursor **doesn't** change on video titles and I have to handle this case explicitly. Can't understand why the original game made this.

About the version info, according to @bgK it is hard-coded in the original executable so we don't have a way to extract it. Based on what I know, we cannot expect that the user can provide the EXE since ResidualVM only needs the game's data files, and the data files are not related to the version info. ResidualVM is not a **part** of TLJ so it feels wired to have its version number in the game, IMO.

PR #1467, opened 2018-06-29
STARK: Improve the main menu

DouglasLiuGamer:
Implement the version info text (#1439) and the Book of Secrets in the main menu.


engines/stark/gfx/renderentry.cpp
```diff
@@ -194,10 +194,16 @@ bool RenderEntry::intersectRay(const Math::Ray &ray) const {
...
 VisualImageXMG *RenderEntry::getImage() const {
+	if (!_visual) {
+		return nullptr;
+	}
 	return _visual->get<VisualImageXMG>();
 }
 
 VisualText *RenderEntry::getText() const {
+	if (!_visual) {
+		return nullptr;
```
DouglasLiuGamer: In the implementation of `GameWindow::onScreenChanged()` I came across with render entries with no `Visual`. It feels strange to me since a `RenderEntry` without a `Visual` seems pretty useless. But anyway it needs to be handled.

engines/stark/services/settings.h
```diff
@@ -85,6 +85,15 @@ class Settings {
 	/** Check whether low-resolution fmv is available */
 	bool hasLowResFMV() { return _hasLowRes; }
 
+	/** Enable the book of secrets */
+	void enableBookOfSecrets() { 
+		ConfMan.setBool("xoBfOsterceS", true);
```
DouglasLiuGamer: Just follow the name in "w_world.ini" :p

engines/stark/services/userinterface.h
```diff
@@ -96,7 +96,9 @@ class UserInterface {
 	void backPrevScreen();
 
 	/** Back to the main menu screen and rest resources */
-	void quitToMainMenu();
+	void requestQuitToMainMenu() { _quitToMainMenu = true; }
+	bool hasQuitToMainMenuRequest() { return _quitToMainMenu; }
+	void performQuitToMainMenu();
```
DouglasLiuGamer: `Script` needs to request quitting to the main menu. In the previous implementation, immediately quitting will trigger the deallocation of the `Script` object and therefore crash the game ~~and it took me an eternity to realize this!!!~~ . So it is cached here and is performed later in the update part of the game's main loop.

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -181,8 +172,24 @@ void MainMenuScreen::settingsHandler() {
 	StarkUserInterface->changeScreen(Screen::kScreenSettingsMenu);
 }
 
+void MainMenuScreen::boxHandler() {
+	if (!StarkSettings->isDemo() && StarkSettings->hasBookOfSecrets()) {
+		StarkUserInterface->changeScreen(kScreenGame);
+		StarkResourceProvider->initGlobal();
+		StarkScene->setFadeLevel(1.0f);
```
DouglasLiuGamer: The "Book of Secrets" will not handle the fade level by itself. I don't find a problem setting it here before the actual game starts.

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -181,8 +172,24 @@ void MainMenuScreen::settingsHandler() {
...
+
 void MainMenuScreen::quitHandler() {
 	StarkUserInterface->notifyShouldExit();
 }
 
-} // End of namespace Stark
\ No newline at end of file
+VersionInfoText::VersionInfoText() :
+		StaticLocationWidget("VERSION INFO REALLY", nullptr, nullptr) {
+	Common::String text = _copyrightSymbol + Common::String("1999 Funcom");
```
DouglasLiuGamer: Since we cannot access the build version of the original executable and we are not using the Audiere library, I chose to just display the Funcom text.

engines/stark/ui/menu/mainmenu.cpp
```diff
@@ -181,8 +172,24 @@ void MainMenuScreen::settingsHandler() {
...
+	_renderEntry->getText()->setText(text);
+	_renderEntry->setPosition(Common::Point(_posX, _posY));
```
DouglasLiuGamer: The default positions of the "VERSION INFO HERE" widget in the released game and the demo are different, and both of them needs to be moved, so I chose to hard-code it.

engines/stark/ui/world/gamewindow.cpp
```diff
@@ -238,4 +241,25 @@ void GameWindow::reset() {
...
+	}
+
+	Common::Array<Resources::Layer *> layers = StarkGlobal->getCurrent()->getLocation()->listLayers();
+
+	VisualText *text = nullptr;
+	Gfx::RenderEntryArray renderEntries;
+	for (uint i = 0; i < layers.size(); ++i) {
+		renderEntries = layers[i]->listRenderEntries();
+		for (uint j = 0; j < renderEntries.size(); ++j) {
+			text = renderEntries[j]->getText();
```
DouglasLiuGamer: Texts in not only the current layer but all the layers within the same location need to reset their textures when the resolution has changed, otherwise, the user may change the screen, go back to a previous layer and find size-unmatched texts.

DouglasLiuGamer:
funcom
A screenshot of the current version info text for reference

DouglasLiuGamer:
A quick way to test the Book of Secrets: Ensure that the "xoBfOsterceS" is not set in the configuration, use the debug console to change to location 3c 00, enable inventory item 15, and plant the organic leaf in the flower bed. The Book of Secrets will then be enabled.

engines/stark/ui/menu/mainmenu.cpp
bgK: Shouldn't the fade level be initialized to 1.0 in the Scene class rather than here?

engines/stark/ui/menu/mainmenu.h
```diff
@@ -55,9 +55,22 @@ class MainMenuScreen : public StaticLocationScreen {
...
+/**
+ * The version info text
+ */
+class VersionInfoText : public StaticLocationWidget {
+public:
+	VersionInfoText();
+	virtual ~VersionInfoText() {}
+
+private:
+	static const int _copyrightSymbol = -87, _posX = 16, _posY = 419;
```
bgK: `_copyrightSymbol` is the ISO-8859-1 code for the copyright symbol. It would be easier to understand what it is as a hexadecimal number (0xA9).

Also please don't put the position on the same line, they are really unrelated.

engines/stark/stark.cpp
```diff
@@ -247,6 +247,11 @@ void StarkEngine::updateDisplayScene() {
 	// Clear the screen
 	_gfx->clearScreen();
 
+	// Quit to main menu screen, if needed
+	if (_userInterface->hasQuitToMainMenuRequest()) {
```
bgK: This check feels like it does not belong here. Couldn't it be in `StarkEngine::mainLoop` with the quit check and the location change check?

engines/stark/ui/world/gamewindow.cpp
```diff
@@ -238,4 +241,25 @@ void GameWindow::reset() {
...
+void GameWindow::onScreenChanged() {
+	// May be called when resources have not been loaded
+	if (!StarkGlobal->getCurrent()) {
+		return;
+	}
+
+	Common::Array<Resources::Layer *> layers = StarkGlobal->getCurrent()->getLocation()->listLayers();
+
+	VisualText *text = nullptr;
+	Gfx::RenderEntryArray renderEntries;
```
bgK: Please reduce the scope of these variables. That'd makes the code a bit easier to read, IMO.

DouglasLiuGamer: Do you mean putting the declaration into the loop? I am not sure how compilers can optimize this these days but isn't that may potentially cause unnecessary variable creations and deallocations, especially since we have an array here?

bgK: C++ compilers are very good with these little details. For non performance critical code it's better to focus on writing simple and readable code.

In this particular case, constructing `RenderEntryArray` outside of the loop may prevent Return Value Optimization.

engines/stark/services/userinterface.h
bgK: Building with Address Sanitizer enabled helps a lot to understand these issues. It tells precisely the cause of the use after free. It's probably a bit of a pain to get it working on Windows, and IMO it's well worth having a Linux install somewhere just to be able to use it when needed.

DouglasLiuGamer:
@bgK If there's nothing left I think we can do the merge.

engines/stark/services/userinterface.h
DouglasLiuGamer: Yeah now I know the good of Linux...

bgK:
Thanks!

PR #1474, opened 2018-07-02
STARK: Implement keyboard bindings

DouglasLiuGamer:
This PR is corresponding to implementing the keyboard bindings in TLJ. It is unfinished, but opened early for review and progress checking.

### DIFF
--- a/engines/stark/resources/item.cpp
+++ b/engines/stark/resources/item.cpp
@@ -175,6 +175,10 @@ void Item::saveLoadCurrent(ResourceSerializer *serializer) {
 	}
 }
 
+Common::Array<Common::Point> Item::listExitPositions() {
+	return Common::Array<Common::Point>();
+}
+
 ItemVisual::~ItemVisual() {
 	delete _renderEntry;
 }
@@ -393,6 +397,28 @@ Common::String ItemVisual::getHotspotTitle(uint32 hotspotIndex) {
 	}
 
 	return title;
+}
+
+Common::Array<Common::Point> ItemVisual::listExitPositionsImpl() {
+	Common::Array<PATTable *> pattables = listChildrenRecursive<PATTable>();
+	
+	Common::Array<Common::Point> positions;
+	Common::Point invalidPosition(-1, -1);
+
+	Common::Array<PATTable *>::iterator element = pattables.begin();
+	while (element != pattables.end()) {
+		if ((*element)->getDefaultAction() == _exitAction) {
+			Common::Point hotspot = getAnim()->getHotspotPosition((*element)->getIndex());
+			if (hotspot != invalidPosition) {
+				hotspot += _renderEntry->getPosition();
+				positions.push_back(hotspot);
+			}
+		}
+
+		++element;
+	}
+
+	return positions;
 }
 
 ItemTemplate::~ItemTemplate() {
@@ -807,6 +833,10 @@ Gfx::RenderEntry *FloorPositionedImageItem::getRenderEntry(const Common::Point &
 	return _renderEntry;
 }
 
+Common::Array<Common::Point> FloorPositionedImageItem::listExitPositions() {
+	return listExitPositionsImpl();
+}
+
 void FloorPositionedImageItem::setPosition2D(const Common::Point &position) {
 	_position = position;
 }
@@ -859,6 +889,10 @@ void ImageItem::printData() {
 
 	debug("reference: %s", _reference.describe().c_str());
 	debug("position: x %d, y %d", _position.x, _position.y);
+}
+
+Common::Array<Common::Point> ImageItem::listExitPositions() {
+	return listExitPositionsImpl();
 }
 
 ModelItem::~ModelItem() {
--- a/engines/stark/resources/item.cpp
+++ b/engines/stark/resources/item.cpp
@@ -405,17 +405,14 @@ Common::Array<Common::Point> ItemVisual::listExitPositionsImpl() {
 	Common::Array<Common::Point> positions;
 	Common::Point invalidPosition(-1, -1);
 
-	Common::Array<PATTable *>::iterator element = pattables.begin();
-	while (element != pattables.end()) {
-		if ((*element)->getDefaultAction() == _exitAction) {
-			Common::Point hotspot = getAnim()->getHotspotPosition((*element)->getIndex());
+	for (uint i = 0; i < pattables.size(); ++i) {
+		if (pattables[i]->getDefaultAction() == _exitAction) {
+			Common::Point hotspot = getAnim()->getHotspotPosition(i);
 			if (hotspot != invalidPosition) {
 				hotspot += _renderEntry->getPosition();
 				positions.push_back(hotspot);
 			}
 		}
-
-		++element;
 	}
 
 	return positions;
--- a/engines/stark/resources/item.cpp
+++ b/engines/stark/resources/item.cpp
@@ -406,7 +406,7 @@ Common::Array<Common::Point> ItemVisual::listExitPositionsImpl() {
 	Common::Point invalidPosition(-1, -1);
 
 	for (uint i = 0; i < pattables.size(); ++i) {
-		if (pattables[i]->getDefaultAction() == _exitAction) {
+		if (pattables[i]->getDefaultAction() == PATTable::kActionExit) {
 			Common::Point hotspot = getAnim()->getHotspotPosition(i);
 			if (hotspot != invalidPosition) {
 				hotspot += _renderEntry->getPosition();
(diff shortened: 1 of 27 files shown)


engines/stark/services/userinterface.cpp
```diff
@@ -396,4 +399,31 @@ void UserInterface::performToggleSubtitle() {
 	_toggleSubtitle = false;
 }
 
+void UserInterface::cycleInventory(int step) {
+	Resources::KnowledgeSet *inventory = StarkGlobal->getInventory();
+	Common::Array<Resources::Item *> inventoryItems = inventory->listChildren<Resources::Item>(Resources::Item::kItemInventory);
+
+	int16 curItem = getSelectedInventoryItem();
+	int16 nextItem = curItem + step;
```
DouglasLiuGamer: The original game has a very strange feature: when you start to cycle the inventory with nothing selected at the beginning, it will skip the first (or the last) item. This happens only in the start and the skipped item will show when you keep on pressing A or S. I can't think of a reasonable way to implement this and it feels like a "bug" to me, so I didn't do it here.

engines/stark/services/userinterface.cpp
```diff
@@ -396,4 +399,31 @@ void UserInterface::performToggleSubtitle() {
...
+		// Boundary cases
+		if (nextItem < -1) {
+			nextItem = inventoryItems.size() - 1;
+		}
+		if (nextItem > 0 && uint(nextItem) >= inventoryItems.size()) {
+			nextItem = -1;
+		}
+
+		// Item None, Hand, Eye, Mouth are skipped
+		if (nextItem < 0 || (inventoryItems[nextItem]->isEnabled() && nextItem > 3)) {
```
DouglasLiuGamer: Can I safely assume that the item None, Hand and Mouth are the first four items in the inventory?

engines/stark/services/userinterface.h
```diff
@@ -165,6 +165,11 @@ class UserInterface {
 	/** Directly open or close a screen */
 	void toggleScreen(Screen::Name screenName);
 
+	/** Toggle subtitles on and off */
+	void requestToggleSubtitle() { _toggleSubtitle = !_toggleSubtitle; }
+	bool hasToggleSubtitleRequest() { return _toggleSubtitle; }
+	void performToggleSubtitle();
```
DouglasLiuGamer: Based on the original game, the toggling of subtitles will be performed after the finishing of the current speech. So some kind of delay needs to be done here.

engines/stark/ui/world/dialogpanel.cpp
```diff
@@ -174,25 +181,44 @@ void DialogPanel::updateDialogOptions() {
...
+	if (!_options.empty()) {
+		_options[_focusedOption]->setActive();
+	}
 }
 
 void DialogPanel::onMouseMove(const Common::Point &pos) {
+	static Common::Point prevPos;
+
+	// Mouse Move event will always exists. Specifically handled here.
+	if (pos == prevPos && !_options.empty()) {
```
DouglasLiuGamer: It seems that the mouse move event will always be there, even when the mouse is not moving at all, yet the dialog panel requires that players can use the up or down key to select options even when the mouse hovers on an option. I am not sure how `onMouseMove()` depends on the event so I only tackle this special case here.

engines/stark/resources/item.cpp
```diff
@@ -405,17 +405,14 @@ Common::Array<Common::Point> ItemVisual::listExitPositionsImpl() {
 	Common::Array<Common::Point> positions;
 	Common::Point invalidPosition(-1, -1);
 
-	Common::Array<PATTable *>::iterator element = pattables.begin();
-	while (element != pattables.end()) {
-		if ((*element)->getDefaultAction() == _exitAction) {
-			Common::Point hotspot = getAnim()->getHotspotPosition((*element)->getIndex());
+	for (uint i = 0; i < pattables.size(); ++i) {
+		if (pattables[i]->getDefaultAction() == _exitAction) {
+			Common::Point hotspot = getAnim()->getHotspotPosition(i);
```
DouglasLiuGamer: What is the usage of `Object::getIndex()` anyway? What's its difference with the index of a listed array like here?

bgK: `Object::getIndex` is used to uniquely identify an entity in the list of children of the parent entity.

In this case, I don't know why the index in the array was used. It seems fragile to me.

engines/stark/stark.cpp
```diff
@@ -202,6 +202,15 @@ void StarkEngine::processEvents() {
 	Common::Event e;
 	while (g_system->getEventManager()->pollEvent(e)) {
 		// Handle any buttons, keys and joystick operations
+
+		if (isPaused()) {
+			// Only pressing key P to resume the game is allowed when the game is paused
+			if (e.type == Common::EVENT_KEYDOWN && e.kbd.keycode == Common::KEYCODE_p) {
+				pauseEngine(false);
+			}
+			continue;
```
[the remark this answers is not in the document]
DouglasLiuGamer: My Steam version will freeze the mouse when the game is paused and I am based on that. Does the game behave the same in other versions?

engines/stark/stark.cpp
```diff
@@ -218,9 +227,103 @@ void StarkEngine::processEvents() {
...
+			} else if (e.kbd.keycode == Common::KEYCODE_F1) {
+				_userInterface->toggleScreen(Screen::kScreenDiaryIndex);
+			} else if (e.kbd.keycode == Common::KEYCODE_F2) {
+				_userInterface->toggleScreen(Screen::kScreenSaveMenu);
+			} else if (e.kbd.keycode == Common::KEYCODE_F3) {
+				_userInterface->toggleScreen(Screen::kScreenLoadMenu);
+			} else if (e.kbd.keycode == Common::KEYCODE_F4) {
+				_userInterface->toggleScreen(Screen::kScreenDialog);
+			} else if (e.kbd.keycode == Common::KEYCODE_F5) {
+				if (_diary->isEnabled()) {
```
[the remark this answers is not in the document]
DouglasLiuGamer: Fixed

DouglasLiuGamer:
The commit 3e38b1c is used to propose a solution for a case like this, assuming that an option may wrap up more than one line.

options

In the original implementation (I mean the one before 2453408), when calculating how many options are displayed on the screen, an option will be added as long as there is remaining space, even though it will exceed it (for 9 pixels, to be specific).

However, for a case shown with the pic above, it will cause problems. If in that case, the `DialogPanel` scrolls down, according to the game's logic, the third option here will become the next "first visible option" and the fifth option here will become the next "last visible option". Yet there is a space of one line left, so the game **should** include one more option, which happens to wrap up two lines, in this case.

In the original implementation, that larger option will be included, which will cause a corner case that the last option will never show up. The game should first check whether the next option may completely fit into the remaining space, then add it to the list. To do so, the original height used to bound the options needs to be enlarged to ensure that all four lines of options can be included.

Actually, I am not sure whether this is how the original game handles this. So many things could happen in scrolling if it is possible to have options wrap up multiple lines. Is it really possible? Is there a location I can check on it in the game?

bgK:
My understanding of the (mostly reverse engineered) initial code is that the game accommodates with options spanning two lines by scrolling options by increments of 3 even if 4 lines can be drawn. That way there can be one option spanning two lines without (too much) trouble.

I've used the French version to look for a dialog option taking two lines as that language is a bit more verbose than English:
before

After scrolling:
after

I don't think it's necessary to implement something too complicated that handles all the cases. Keeping the status-quo is fine with me.

DouglasLiuGamer:
> The game accommodates with options spanning two lines by scrolling options by increments of 3 even if 4 lines can be drawn.

This sounds like what the old implementation does, yet that implementation behaves differently with my Steam version. In the old implementation, if you scroll down to the bottom you'll have only 3 options displayed:

old

But in the Steam version, you should have a full four of them:

game

What did you mean by "status-quo"? The original "scrolling by 3" or the new one I proposed? The one I wrote is not very complicated, just some very basic checking, and it doesn't seem to decrease the performance. If there is no bug there I prefer to use the one I propose.

DouglasLiuGamer:
@bgK If you think everything is good, I think it can be merged.

bgK:
I noticed that with this branch the dialog speech is not working anymore -- the dialogs are entirely silent. Can you take a look ?

DouglasLiuGamer:
@bgK That's strange, dialogs work fine with speeches on my computer. Where did you get a silent dialog?

bgK:
Indeed, something I don't understand is happening. I was able to reproduce the silent dialogs twice tonight, but not anymore after that. There's no memory problems detected by Valgrind / ASan so I'll say it was a mistake on my end for now.

Thanks for your work; merging.

PR #1482, opened 2018-07-25
SDL: Enhance warning/error output

raziel-:
I think it would be a good idea to enhance the warning/error output for the last three SDL driven error/warning messages.

I got two instances where it threw an error which looked like this on my console: "Error: Invalid window" and i was unable to tell where the error came from. Searching it in the code brought me to SDL_GetError.
So, imho it would be a good thing to have that information printed as well.

I tried to stay conform to the SDL command that invoked the error/warning, but i might still be wrong.
If there are objections or parts i need to change, please voice up


bgK:
Thanks!

PR #1483, opened 2018-07-25
STARK: Implement the shadow

DouglasLiuGamer:
This PR is corresponding to implementing the characters' shadows in TLJ.

### DIFF
--- a/engines/stark/gfx/openglsactor.cpp
+++ b/engines/stark/gfx/openglsactor.cpp
@@ -361,7 +361,7 @@ void OpenGLSActorRenderer::setShadowUniform(const LightEntryArray &lights, const
 }
 
 bool OpenGLSActorRenderer::getPointLightContribution(LightEntry *light,
-		const Math::Vector3d &actorPosition, Math::Vector3d &direction) {
+		const Math::Vector3d &actorPosition, Math::Vector3d &direction, float weight) {
 	float distance = light->position.getDistanceTo(actorPosition);
 
 	if (distance > light->falloffFar) {
@@ -387,7 +387,7 @@ bool OpenGLSActorRenderer::getPointLightContribution(LightEntry *light,
 
 	direction = actorPosition - light->position;
 	direction.normalize();
-	direction *= factor * brightness;
+	direction *= factor * brightness * weight;
 
 	return true;
 }
@@ -407,10 +407,10 @@ bool OpenGLSActorRenderer::getDirectionalLightContribution(LightEntry *light,
 }
 bool OpenGLSActorRenderer::getSpotLightContribution(LightEntry *light,
 		const Math::Vector3d &actorPosition, Math::Vector3d &direction) {
-	Math::Vector3d LightToActor = actorPosition - light->position;
-	LightToActor.normalize();
+	Math::Vector3d lightToActor = actorPosition - light->position;
+	lightToActor.normalize();
 
-	float cosAngle = MAX(0.0f, LightToActor.dotProduct(light->direction));
+	float cosAngle = MAX(0.0f, lightToActor.dotProduct(light->direction));
 	float cone = (cosAngle - light->innerConeAngle.getCosine()) /
 			MAX(0.001f, light->outerConeAngle.getCosine() - light->innerConeAngle.getCosine());
 	cone = CLIP(cone, 0.0f, 1.0f);
@@ -419,7 +419,7 @@ bool OpenGLSActorRenderer::getSpotLightContribution(LightEntry *light,
 		return false;
 	}
 
-	return getPointLightContribution(light, actorPosition, direction);
+	return getPointLightContribution(light, actorPosition, direction, cone);
 }
 
 
--- a/engines/stark/gfx/openglsactor.cpp
+++ b/engines/stark/gfx/openglsactor.cpp
@@ -116,6 +116,9 @@ void OpenGLSActorRenderer::render(const Math::Vector3d &position, float directio
 	_shader->unbind();
 
 	// Shadow
+	glEnable(GL_BLEND);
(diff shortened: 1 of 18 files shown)


DouglasLiuGamer:
A screenshot of the current version. The shadow is directly cast from above.

shadow

engines/stark/gfx/openglsactor.cpp
```diff
@@ -112,6 +116,34 @@ void OpenGLSActorRenderer::render(const Math::Vector3d &position, float directio
 	}
 
 	_shader->unbind();
+
+	if (castsShadow && StarkSettings->getBoolSetting(Settings::kShadow)) {
+		glEnable(GL_BLEND);
+		glEnable(GL_STENCIL_TEST);
+
+		_shadowShader->enableVertexAttribute("position1", _faceVBO, 3, GL_FLOAT, GL_FALSE, 14 * sizeof(float), 0);
```
Botje: Although you're just emulating code that is already written, I would like to remark two things:
1. Vertex attributes should only be set once, when the corresponding VBO object is created, rather than for every render operation.
2. Both the stride (second to last parameter) and the offset (last parameter) should be defined in terms of multiples of floats.

engines/stark/gfx/openglsactor.cpp
```diff
@@ -112,6 +116,34 @@ void OpenGLSActorRenderer::render(const Math::Vector3d &position, float directio
 	}
 
 	_shader->unbind();
+
+	if (castsShadow && StarkSettings->getBoolSetting(Settings::kShadow)) {
+		glEnable(GL_BLEND);
+		glEnable(GL_STENCIL_TEST);
```
bgK: I'm not under the impression the stencil buffer is actually used.

engines/stark/shaders/stark_shadow.fragment
```diff
@@ -0,0 +1,5 @@
+OUTPUT
+
+void main() {
+	outColor = vec4(0.0, 0.0, 0.0, 0.6);
```
bgK: Isn't the shadow a bit too dark ?

engines/stark/gfx/openglsactor.cpp
Botje: See the `glStencilFunc` and `glStencilOp` calls in `set3DMode`. The first makes it so that every touched pixel ends with a stencil value of 1 (or more if overdraw occurs), while the second discards pixels that have a non-zero stencil value. This prevents the shadows from overlapping and creates a uniform shadow shape.

engines/stark/gfx/openglsactor.cpp
```diff
@@ -298,5 +304,125 @@ void OpenGLSActorRenderer::setLightArrayUniform(const char *uniform, const Light
...
+	} else {
+		// Cast from above by default
+		sumDirection.x() = 0;
+		sumDirection.y() = 0;
+		sumDirection.z() = -1;
+	}
+
+	//Transform the direction to the model space and pass to the shader
+	sumDirection = worldToModelRot * sumDirection;
+	_shadowShader->setUniform("lightDirection", sumDirection);
```
DouglasLiuGamer: Rather than computing a matrix for casting shadows for vertices in general, I think it is a lot easier and intuitive to transform the light direction to the model space and then do the projection in the shader.

DouglasLiuGamer:
With the tremendous help from @bgK who helped me to look at the disassembly, now the shadow is computed based on the actual lights. Some screenshots of the current version here, I temporarily rendered the shadow in solid green for better illustration.

longshadow1

longshadow2

In location 39 00 the shadow will get long and change direction when April moves.

I'll improve the code based on the above review soon.

engines/stark/gfx/openglsactor.cpp
```diff
@@ -266,5 +304,125 @@ void OpenGLSActorRenderer::setLightArrayUniform(const char *uniform, const Light
...
+	float cosAngle = MAX(0.0f, LightToActor.dotProduct(light->direction));
+	float cone = (cosAngle - light->innerConeAngle.getCosine()) /
+			MAX(0.001f, light->outerConeAngle.getCosine() - light->innerConeAngle.getCosine());
+	cone = CLIP(cone, 0.0f, 1.0f);
+
+	if (cone <= 0) {
+		return false;
+	}
+
+	return getPointLightContribution(light, actorPosition, direction);
```
bgK: `cone` should affect `direction`.

engines/stark/gfx/openglsactor.cpp
```diff
@@ -266,5 +304,125 @@ void OpenGLSActorRenderer::setLightArrayUniform(const char *uniform, const Light
...
+
+	direction = light->direction;
+	direction.normalize();
+	direction *= brightness;
+
+	return true;
+}
+bool OpenGLSActorRenderer::getSpotLightContribution(LightEntry *light,
+		const Math::Vector3d &actorPosition, Math::Vector3d &direction) {
+	Math::Vector3d LightToActor = actorPosition - light->position;
```
bgK: That variable name should be lowerCamelCase.

DouglasLiuGamer:
@Botje About the two remarks you made, I am not sure about the first one. Based on the current logic in the rendering code, the same shader is used for different models, which has there own VBO. Therefore it should be natural to bind the VBO to vertex attributes everytime the model is drawn, at least this is what I think about it.

Well, I am new to OpenGL, so I may be wrong.

DouglasLiuGamer:
@bgK I can't reproduce the case in your video. The shadow on my computer there changes seamlessly. Did you use the latest version where the `cone` of the spotlight affects the direction?

DouglasLiuGamer:
~~I felt so stupid for my last commit~~
@bgK If there is nothing more, I think we can do the merging.

PR #1484, opened 2018-07-30
Fixed the changechapter problem

### LINKED ISSUE #1466: STARK: Setting chapter with "changeChapter" debug command very high (or low) results in weird chapter values
Related to #1456.

There seems to be no checks performed on setting the value of the "setChapter" command – while luckily it doesn't crash the game, setting too high (or too low) value results in wrong values being set, for example bu setting `changeChapter -9999999999` I get the value of -14100565407 (opposite in case of the positive values).

It seems to happen starting at 2^31 so `changeChapter 2147483647` is okay, while `changeChapter 2147483648` overflows the value to -2147483648. Then setting `changeChapter 2147483649` continues to go higher setting it to -2147483647

Setting it to even more extreme value, such as `changeChapter 99999999999999999999` sets the chapter at -1. I didn't tested at which point exactly does it happen.

orangeforest11:
Fixes #1466

### DIFF
--- a/engines/stark/console.cpp
+++ b/engines/stark/console.cpp
@@ -38,7 +38,7 @@
 #include "engines/stark/services/services.h"
 #include "engines/stark/services/staticprovider.h"
 #include "engines/stark/tools/decompiler.h"
-
+#include "limits.h"
 #include "common/file.h"
 
 namespace Stark {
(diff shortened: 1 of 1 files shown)


bgK:
thanks!

DouglasLiuGamer:
Wouldn't it be a lot better if we check whether the value is within the true range of the game? If I remember correctly we only have, like, 13 chapters? So the maximum number will be only 130.

PR #1485, opened 2018-08-02
STARK: Add range checking in acquiring chapters' titles and subtitles

### LINKED ISSUE #1466: STARK: Setting chapter with "changeChapter" debug command very high (or low) results in weird chapter values

DouglasLiuGamer:
Fix #1466

### DIFF
--- a/engines/stark/services/gamechapter.cpp
+++ b/engines/stark/services/gamechapter.cpp
@@ -29,7 +29,7 @@
 
 namespace Stark {
 
-GameChapter::GameChapter() {
+GameChapter::GameChapter() : _errorText("Unknown Chapter") {
 	Common::INIFile file;
 	if (!file.loadFromFile("chapters.ini")) {
 		error("Opening file 'chapters.ini' failed");
@@ -56,7 +56,7 @@ GameChapter::GameChapter() {
 		key = Common::String::format("%02d", index);
 	}
 
-	if (index < 15) {
+	if (index < _numChapter) {
 		error("File 'chapters.ini' is incomplete");
 	}
 }
--- a/engines/stark/services/gamechapter.h
+++ b/engines/stark/services/gamechapter.h
@@ -39,24 +39,35 @@ public:
 	~GameChapter() {}
 
 	const Common::String &getChapterTitle(uint chapter) const {
-		return _chapterEntries[chapter / 10].title;
+		if (chapter >= _numChapter * 10) {
+			return  _errorText;
+		} else {
+			return _chapterEntries[chapter / 10].title;
+		}
 	}
 
 	const Common::String &getCurrentChapterTitle() const;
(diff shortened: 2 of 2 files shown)


orangeforest11:
Isn't it better to not hardcode the 150 value?

DouglasLiuGamer:
@orangeforest11 The only place I can think of to get the 150 value is through the chapter.ini file. Yet if the number of chapters contained by that file is less than 15, the game will not run (At least this is what the current logic does). Since the game is already fixed, I personally don't see the need to dynamically get the value.

engines/stark/console.cpp
```diff
@@ -588,6 +587,8 @@ bool Console::Cmd_ChangeLocation(int argc, const char **argv) {
 }
 
 bool Console::Cmd_ChangeChapter(int argc, const char **argv) {
+	static const int32 maxChapterValue = 150;
```
bgK: Other chapter checks use 999 as the max value. I see no reason to be very restrictive here, console commands are for developers to debug the game.

engines/stark/console.cpp
```diff
@@ -599,12 +600,13 @@ bool Console::Cmd_ChangeChapter(int argc, const char **argv) {
...
 		return true;
 	}
-
-	long value = strtol(argv[1], 0, 10);
-	if (value >= 0 && value <= INT_MAX)
-		StarkGlobal->setCurrentChapter((int32) value);
-	else
-		debugPrintf("Invalid chapter\n");
+	
+	long value = strtol(argv[1], nullptr, 10);
```
bgK: Actually the check is not working because `strtol` is not called correctly. I can say `changeChapter chapter1` and get no error.

DouglasLiuGamer:
The branch in this PR was rebased once.

DouglasLiuGamer:
The branch was rebased once.

@bgK If that's the case, I chose to add a checking in `GameChapter` service, so that at least an invalid chapter value will not crash the game.
About the input checking, I think it is kind of meaningless to make it fancy since it is just a debugging tool for developers. Maybe keeping the status quo is enough.

bgK:
Thanks!

PR #1488, opened 2018-08-07
STARK: Enabled and fixed action menu and cursor scaling

orangeforest11:
No reason not to scale them. As it currently is the proportions of these things are messed up when using high resolution, harming the user experience, which of course doesn't happen with the original game where there are only two possibilities: either no scaling at all and everything has the right proportions, or everything is scaled (by the monitor, GPU, or otherwise as the original game doesn't support scaling at all) with everything still having the right proportions. But there is no possibility in the original game where the situation is like it currently is with ResidualVM, where most things are scaled, but cursor and action menu aren't.

### DIFF
--- a/engines/stark/gfx/driver.cpp
+++ b/engines/stark/gfx/driver.cpp
@@ -69,7 +69,7 @@ void Driver::toggleFullscreen() const {
 void Driver::computeScreenViewport() {
 	int32 screenWidth = g_system->getWidth();
 	int32 screenHeight = g_system->getHeight();
-	
+
 	if (g_system->getFeatureState(OSystem::kFeatureAspectRatioCorrection)) {
 		// Aspect ratio correction
 		int32 viewportWidth = MIN<int32>(screenWidth, screenHeight * kOriginalWidth / kOriginalHeight);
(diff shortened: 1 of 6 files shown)


bgK:
Thanks for your contribution. I see two main issues that need to be fixed before this can be merged.
* The mouse cursor position is now affected by scaling. It's no longer possible to move the mouse by one native pixel, resulting in less fluid mouse movement.
* The mouse cursor is no longer visible on top of the side black borders. The black borders appear to be sticky from the user perspective.

engines/stark/gfx/driver.cpp
```diff
@@ -69,7 +69,7 @@ void Driver::toggleFullscreen() const {
 void Driver::computeScreenViewport() {
 	int32 screenWidth = g_system->getWidth();
 	int32 screenHeight = g_system->getHeight();
-
+
```
bgK: Please make sure you don't leave trailing spaces behind.

engines/stark/gfx/driver.cpp
```diff
@@ -85,13 +85,27 @@ void Driver::computeScreenViewport() {
...
-Common::Rect Driver::gameViewport() const {
-	Common::Rect game = Common::Rect(_screenViewport.width(), _screenViewport.height() * kGameViewportHeight / kOriginalHeight);
-	game.translate(_screenViewport.left, _screenViewport.top + _screenViewport.height() * kTopBorderHeight / kOriginalHeight);
-
+Common::Rect Driver::gameViewport(bool unscaled) const {
+	uint32 width = 640;
+	uint32 height = 480;
+	uint32 left = 0;
+	
+	if(unscaled){
```
bgK: Please make sure to follow our code formatting conventions: http://wiki.scummvm.org/index.php/Code_Formatting_Conventions

bgK:
Thanks, I've merged this manually with some more changes.

PR #1489, opened 2018-08-11
STARK: Set default keyframe for bone animation

### LINKED ISSUE #1436: STARK: Model of old woman in the location forest is glitching
old woman

Save game from GOG version: (April must enter "Into forest" exit):
[tlj-win-027.zip](https://github.com/residualvm/residualvm/files/2066360/tlj-win-027.zip)

ResidualVM console:

> WARNING: Unable to animate bone '1' at 4988 ms!
WARNING: Unable to animate bone '2' at 4988 ms!
WARNING: Unable to animate bone '3' at 4988 ms!
WARNING: Unable to animate bone '4' at 4988 ms!
WARNING: Unable to animate bone '5' at 4988 ms!
WARNING: Unable to animate bone '6' at 4988 ms!
WARNING: Unable to animate bone '7' at 4988 ms!
WARNING: Unable to animate bone '9' at 4988 ms!
WARNING: Unable to animate bone '10' at 4988 ms!
WARNING: Unable to animate bone '11' at 4988 ms!
WARNING: Unable to animate bone '12' at 4988 ms!
WARNING: Unable to animate bone '13' at 4988 ms!
WARNING: Unable to animate bone '15' at 4988 ms!
WARNING: Unable to animate bone '17' at 4988 ms!
WARNING: Unable to animate bone '18' at 4988 ms!
WARNING: Unable to animate bone '19' at 4988 ms!
WARNING: Unable to animate bone '20' at 4988 ms!
WARNING: Unable to animate bone '21' at 4988 ms!
WARNING: Unable to animate bone '23' at 4988 ms!
WARNING: Unable to animate bone '25' at 4988 ms!
WARNING: Unable to animate bone '26' at 4988 ms!
WARNING: Unable to animate bone '27' at 4988 ms!
WARNING: Unable to animate bone '28' at 4988 ms!
WARNING: Unable to animate bone '30' at 4988 ms!
WARNING: Unable to animate bone '31' at 4988 ms!
WARNING: Unable to animate bone '32' at 4988 ms!
WARNING: Unable to animate bone '33' at 4988 ms!

orangeforest11:
Mitigates #1436. Glitches were appearing at old lady in the forest because wasn't getting a key frame

Now, if no key frame found, defaults to last key frame with warning message.

### DIFF
--- a/engines/stark/model/skeleton_anim.cpp
+++ b/engines/stark/model/skeleton_anim.cpp
@@ -77,15 +77,9 @@ void SkeletonAnim::getCoordForBone(uint32 time, int boneIdx, Math::Vector3d &pos
 
 		return;
 	}
-
+	
 	for (Common::Array<AnimKey>::const_iterator it = keys.begin(); it < keys.end(); ++it) {
-		if (it->_time == time) {
-			const AnimKey *key = it;
-			pos = key->_pos;
-			rot = key->_rot;
-
-			return;
-		} else if (it->_time > time) {
+		if (it->_time > time) {
 			// Between two key frames, interpolate
 			const AnimKey *a = it;
 			--it;
@@ -97,9 +91,16 @@ void SkeletonAnim::getCoordForBone(uint32 time, int boneIdx, Math::Vector3d &pos
 			rot = b->_rot.slerpQuat(a->_rot, t);
 
 			return;
+		}else if (it->_time == time || it == keys.end() - 1){
+			// At a key frame or if none found
+			const AnimKey *key = it;
+			pos = key->_pos;
+			rot = key->_rot;
+
+			return;
 		}
 	}
-
+	
 	warning("Unable to animate bone '%d' at %d ms", boneIdx, time);
 }
 
--- a/engines/stark/model/skeleton_anim.cpp
+++ b/engines/stark/model/skeleton_anim.cpp
@@ -77,7 +77,7 @@ void SkeletonAnim::getCoordForBone(uint32 time, int boneIdx, Math::Vector3d &pos
 
 		return;
 	}
-	
+
 	for (Common::Array<AnimKey>::const_iterator it = keys.begin(); it < keys.end(); ++it) {
 		if (it->_time > time) {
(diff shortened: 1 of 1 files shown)


bgK:
I'm sorry, but this solution seems like it's only hiding the effects of the root problem. The animation system should not request out of bounds values in the first place.

orangeforest11:
In case no key frame is found, I think it's better for it to default to some keyframe, rather than the current behavior which might trigger some users visually (with epilepsy?) I guess. Let's take an example of a video player and you're playing a video file and 500 frames of it (5 seconds) are missing for some part of the video, should it a) default to the previous frame so image freezes for 5 seconds, or b) start doing a crazy sequence of random lights that may give you a seizure? Of course it's better to get to the root of the problem, but there should be some reasonable default because nothing will ever be 100% perfect. What even is going on with the memory in case there's no key frame? Why does it default to a series of arbitrary changing values? Are there memory problems/buffer overflows/something?

But I don't see why there would be a problem with the animation system. The problem is either a) there's a problem with this specific animation data file, or b) there's a problem with the animation data reading mechanism. For the second case I guess there might be an unknown version of the animation data file format, or unknown variables/values within it? But why would this happen for only one model?

Perhaps the original game relied on a hack, or different internal coding, for it to work with this model. Maybe the original file isn't correctly formatted or corrupted (or corrupted in some versions of the game, such as the one the example savegame was using?). Yet still, in any case, there should be a reasonable default.

bgK:
It's ok to have a saner default behavior in case no animation frames are found, but it's definitely not ok to disable the warning that was here to warn users and developers that there is a problem that needs to be fixed. It's also not ok to claim to fix the bug, if this is just hiding the most visible effects.

> But is it really a problem with the animation system?
I'm pretty sure there is. It seems to me the animation system is using the animation data of the animation `Sit Idle` (duration 2000 ms) but the animation length of `Sit hurting` (5000 ms).

orangeforest11:
Warning message added and original comment adjusted.

bgK:
Thanks !

PR #1518, opened 2019-02-01
GRIM: Image format must match internalformat for sprites too (ES2)

usineur:
"swapRandB" variable must be set for sprites too, when ResidualVM is built with opengles2 + shaders support.

It fixes this kind of wrong colors rendered:

Here, flames color:
bad1  good1

Other example, ship and blaze colors:
bad2
good2


bgK:
Thanks !

Issue #1411, opened 2018-04-22
SDL: Improve mapping of game controllers in GRIME


lmerckx:
Since the game controller support in ResidualVM, some keys are not available anymore.

For instance, the "skip scene" is now mapped over SDL_CONTROLLER_BUTTON_GUIDE (which is not always present). The menu is associated to SDL_CONTROLLER_BUTTON_RIGHTSTICK (not very intuitive).

Can we improve that ?
My proposition is:
- support "special" buttons directly in resvm-sdl-events.cpp (example: open virtual keyboard)
- in engine classes (grim.cpp or grim_controls.cpp): map SDL2 controller button codes to correct KEY_JOY1_B? value. This will need to add new event types: EVENT_CONTROLLERAXIS_MOTION, EVENT_CONTROLLERBUTTON_DOWN, ... 

What do you think about it ?

lmerckx:
I can do the dev for this issue. But my intention was to share the proposition to get feedback.

bgK:
Sorry for the late reply.

A few years ago a "keymapper" was added to ScummVM / ResidualVM. It's unfinished, and currently not very usable. But the ultimate goal, IMO, is to have things working in the following way:
* Each engine exposes a list of possible actions for each game (Menu, take item, use, skip scene, ...), and default bindings for keyboard and mouse.
* Each backend declares a list of available input devices / keys.
* Each backend has the possibility to override / add to the default bindings for each game.
* A GUI dialog allows the user to customize the action / key bindings.

This is a long term goal. But perhaps one to keep in mind when designing short term solutions so that they don't go in too conflicting directions.

Your proposal seems good to me. Bonus points if you can avoid duplicating the special key handing between `sdl-events.cpp` and `resvm-sdl-events.cpp`.

lmerckx:
@bgK Ok, I nearly developped the above proposition.
But, I still have an issue with the virtual keyboard:
Indeed, the "resvm-sdl-events" needs to know if virtual keyboard is opened to emulate mouse events instead of joystick one.
But it is currently not accessible outside "default-events" which creates it.

Do you think it is acceptable to convert VirtualKeyboard as a singleton with global object (g_vkeybd) ?
It works this way, but I should propose this modification to ScummVM. Will it be accepted ?

lmerckx:
@bgK Or maybe I can test "CursorMan.isVisible()" which indicates (for GrimE games) that the virtual keyboard is opened ?
Very easier solution and ResidualVM only.

[...]

PR #1520, opened 2019-02-06
Improvement of game controller support for Grim engine

lmerckx:
Associated to issue https://github.com/residualvm/residualvm/issues/1411

As explained in the issue:
- Use back button to open virtual keyboard (managed by resvm-sdl-events)
- Map controller buttons to LUA buttons (in grim engine)
- Solve bug on RPI with Myst3 (camera stuck on roof when using joystick)

Tested on RPI2 and X64 (Linux) with Grim, EMI and Myst3


backends/events/sdl/resvm-sdl-events.cpp
```diff
@@ -89,7 +103,8 @@ bool ResVmSdlEventSource::shouldGenerateMouseEvents() {
 		return true;
 	}
 
-	if (g_gui.isActive()) {
+	// FIXME: call to CursorMan.isVisible is not clean ! But don't know how to do better
+	if (g_gui.isActive() || CursorMan.isVisible()) {
```
bgK: Probably checking if the engine is paused is a little better. In any case there needs to be a comment here explaining that this is to send mouse events when the virtual keyboard is open. It's not obvious at all.

backends/events/sdl/resvm-sdl-events.cpp
```diff
@@ -115,7 +130,7 @@ bool ResVmSdlEventSource::handleKbdMouse(Common::Event &event) {
 			_km.x = oldKmX;
 			_km.y = oldKmY;
 		}
-
+		else
```
bgK: What does this change fix?

common/events.h
```diff
@@ -91,7 +91,10 @@ enum EventType {
 	,
 	EVENT_JOYAXIS_MOTION = 23,
 	EVENT_JOYBUTTON_DOWN = 24,
-	EVENT_JOYBUTTON_UP = 25
+	EVENT_JOYBUTTON_UP = 25,
+	EVENT_CONTROLLERAXIS_MOTION = 26,
```
bgK: Joystick and Controller events being something different is a quirk of SDL2. It would be much better if ResidualVM did not inherit the same flaw. Can we have an enum in ResidualVM defining what the joystick buttons are (based on the SDL game controller buttons) and in `ResVmSdlEventSource` something that tries to convert the old joystick button numbers into that enum ?

backends/events/sdl/resvm-sdl-events.cpp
lmerckx: Good idea ... I will check in that way !
Thank you.

backends/events/sdl/resvm-sdl-events.cpp
lmerckx: Without it, the camera becomes crazy when moving the joystick in Myst3 (on RPI) and it directly points to the zenith !
I think it's linked to a bug of warpMouse in SDL2 (2.0.8 and lower versions).

I'll recheck and put a descriptive comment.

common/events.h
lmerckx: The objective was to let the engine choose what to do, but I agree with you, it must be simplified.
Ok, I will try something like that.

backends/events/sdl/resvm-sdl-events.cpp
usineur: I can confirm this change fixes the same behavior in my WIP Nintendo Switch port.

bgK: Ok, it makes sense to me now. I suspect SDL generates a mouse moved event when `warpMouseInWindow` is called.

Can you please write the code as:
```cpp
if (graphicsManager) {
    if (graphicsManager->isMouseLocked()) {
       _km.x = oldKmX;{
       _km.y = oldKmY;
    } else {
        graphicsManager->getWindow()->warpMouseInWindow((Uint16)(_km.x / MULTIPLIER), (Uint16)(_km.y / MULTIPLIER));
    }
}
```

lmerckx:
@bgK I just finished some modifications following your remarks:
* change the test on cursor by a test on engine paused + add comments
* "else" code re-formatted + add comments
* rewrite and simplify Grim engine joystick events (using controller events instead of joystick)

I also added a support of L2/R2 buttons/triggers: allow to have more available buttons + needed for PS2

Is it better for you ?

bgK:
Thanks. I've merged this manually with some little changes.

PR #1525, opened 2019-02-13
SDL: Add the support of saving screenshots

DouglasLiuGamer:
These are works referring to what has been done in ScummVM. The hotkey is the same, `Alt + s`.

The key implementation of creating the screenshots is basically the same as ScummVM, I simply make it adjustable with the existed structure of ResidualVM.

I also connect the screenshot saving feature to the F8 key in *The Longest Journey*, which is used by the original game. Yet to achieve this I decide to modify the `sdl.h` file, which comes from ScummVM. I hope this is okay since I see some functions there marked "ResidualVM specific".

I also update the `README.md` to include this new feature. The configuration part is also based on ScummVM's.

### DIFF
--- a/engines/stark/services/userinterface.cpp
+++ b/engines/stark/services/userinterface.cpp
@@ -58,6 +58,9 @@
 
 #include "gui/message.h"
 
+#include "backends/platform/sdl/sdl.h"
+#include "backends/graphics/sdl/resvm-sdl-graphics.h"
+
 namespace Stark {
 
 UserInterface::UserInterface(Gfx::Driver *gfx) :
@@ -489,7 +492,13 @@ void UserInterface::handleKeyPress(const Common::KeyState &keyState) {
 	} else if (keyState.keycode == Common::KEYCODE_F7) {
 		toggleScreen(Screen::kScreenSettingsMenu);
 	} else if (keyState.keycode == Common::KEYCODE_F8) {
-		warning("TODO: Implement the screenshot saving to local game directory");
+		OSystem_SDL *g_systemSdl = dynamic_cast<OSystem_SDL*>(g_system);
+		if (g_systemSdl) {
+			ResVmSdlGraphicsManager *graphicsMan = dynamic_cast<ResVmSdlGraphicsManager*>(g_systemSdl->getGraphicsManager());
+			if (graphicsMan) {
+				graphicsMan->saveScreenshot();
+			}
+		}
 	} else if (keyState.keycode == Common::KEYCODE_F9) {
 		if (isInGameScreen()) {
 			_shouldToggleSubtitle = !_shouldToggleSubtitle;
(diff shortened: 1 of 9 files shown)


DouglasLiuGamer:
@bgK 

Things on my Mac behaves differently from what you described. OpenGL screen capturing gives all black on Grim and EMI. I have made the commit to fix this. At least in all my testings, the whole window, including the GMM / debug console, will be captured in all cases.

As for the endianness problem, I am not sure about it. The only place where ScummVM considers endianness is in the call `SDL_CreateRGBSurface(...)`, which is trivial. My understanding is that as long as the bytes of color components are stored in RGB order, everything is fine since we are not reading or writing them as a whole `int` here. Plus `Graphics::PixelFormat(3, 8, 8, 8, 0, 16, 8, 0, 0)` is the required pixel format for `Image::writePNG(...)` to work. I don't have a big-endianness working machine ~(man, do we have them nowadays?)~ so I am not sure about what will the problem be...

bgK:
Thanks for the fixes.

Regarding the endianness:
OpenGL returns the pixels in byte order `RGB`, which means the bytes have the `RGB` layout on all systems.
`Graphics::PixelFormat(3, 8, 8, 8, 0, 16, 8, 0, 0)` constructs words of pixel data using this expression `r << 16 | g << 8 | b`. Which means the bytes should be stored as `BGR` on LE systems and `RGB` on BE systems.
But then the PNG code uses `PNG_COLOR_TYPE_RGB` which expects pixel data to be in `RGB` byte order. So in the end the whole thing appears to work by accident with the pixel format being inconsistent with the actual pixel data. So yeah, fixing that mess is a task for another day.

If it's ok with you I will merge this PR shortly.

DouglasLiuGamer:
OK! Thanks for the review :)

engines/stark/services/userinterface.cpp
```diff
@@ -489,7 +492,13 @@ void UserInterface::handleKeyPress(const Common::KeyState &keyState) {
 	} else if (keyState.keycode == Common::KEYCODE_F7) {
 		toggleScreen(Screen::kScreenSettingsMenu);
 	} else if (keyState.keycode == Common::KEYCODE_F8) {
-		warning("TODO: Implement the screenshot saving to local game directory");
+		OSystem_SDL *g_systemSdl = dynamic_cast<OSystem_SDL*>(g_system);
```
bgK: Sorry, I missed that code. You really shouldn't call the SDL backend like this here. Our Android backend does not use SDL, and we can add more non-SDL backends in the future. The proper way to do that is to add a `saveScreenshot` to the `OSystem` class that backends can implement.

bgK:
I've merged this manually.

DouglasLiuGamer:
@bgK 

Sorry for being this late. Too busy last week.

Adding a `saveScreenshot()` to `OSystem` was one of my original plans, but I want to modify as few codes from ScummVM as possible, working on `OSystem` seems to big a move too big.

Also, I think that since now it is the graphics manager that responds to the `Alt + S` event, doesn’t it feel kind of strange to have such an interface on `OSystem`? And now it is only used in TLJ. Personally I think the logic would be better if we also move the listening of the screenshot event to somewhere more close to `OSystem`.

But anyway, as long as you think it is fine, I am okay with the merge.

PR #1533, opened 2019-04-01
Fix mouse going out of bounds in fullscreen

orgads:
The coordinates were translated, but this was done on a copy
instead of overwriting the original object.

In ScummVM this is a reference.

Found the bug with grim-mouse fork.


bgK:
Good catch!

Issue #1451, opened 2018-06-19
EMI: GOG Version is not recognized (Fixed with workaround, details/steps inside)


jasonridesabike:
Just picked up the GOG version of MI4 (released today, [link here](https://www.gog.com/game/escape_from_monkey_island)) and attempted to run it with ResidualVM with no success. Attempting to add the game returns the error message: 

"ScummVM could not find any game in the specified directory!"

File structure appears to generally adhere to requirements listed [here](http://wiki.residualvm.org/index.php?title=Datafiles#Escape_From_Monkey_Island) with the exception that it appears that the texture file FullMonkeyMap.imt appears to have been combined from both cd's into a single file.

Let me know if you need testing

UPDATE: I was running version 3.1. Updating to the daily build allows the game to be detected and copying FullMonkeyMap.imt into 2 files and renaming each respectively to FullMonkeyMap1.imt and FullMonkeyMap2.imt get the game to load. Will create a new issue if I find any problems. 

Thanks!

MrLint:
Thanks for the info. I got it to load with the file copying, but there seems to be issues with some textures not loading even early on. near the very beginning timmy the monkey and the catapult have no textures. (may be more just started playing)

sigboe:
Hey, I am guessing that the game actually ships with the patch integrated. I can provide detailed information on files if that would help. Just give me a hint what to check for.

So we know that all the textures are combined, so it should be possible to parse everything from the one file, and just check the md5 or sha1 sum of the file?

md5sum
    cd9afa776a570152eea5554a24652b41  FullMonkeyMap.imt

    c627acdfa4bbd1a84f7ce4a684cd3d0d  Monkey4.exe
    0e5c6336348e4f831473f6aa89a41040  Monkey.exe

Botje:
We need the md5sum of the first 5000 bytes as well for the game detection. I'll take a look at the contents of FullMonkeyMap on the GOG version next time I have some time.

sigboe:
@Botje Also posting my command so you can see if I did wrong

    dd if=FullMonkeyMap.imt bs=1 count=5000 2>/dev/null | md5sum
    e351d0f1df48f69861d4b2b6af1a0735  -

Botje:
Downloading MonkeyUpdate.exe and putting it in the EMI directory (as instructed by ResidualVM) works for me with the current daily build.

johnsout:
Hi,
Do we have any update on this?

PR #1542, opened 2019-05-12
EMI: Support merged FullMonkeyMap.int as well

### LINKED ISSUE #1541: EMI: No Music On Steam Version
The steam version of the game only includes 1 FullMonkeyMap.imt. I have tried replacing the single file with the two from the cds and renaming them.as instructed. I have also tried manually splitting the fullmonkeymap file included with steam into two files as it seems to be just the two put together into one. Still no music. I am not receiving any error messages. I am running this on retropie through a raspberry pi 3 b.

Botje:
Fixes #1541. Maybe #1451 too?

I reworked the load logic to just load all three known files in succession and then check if the number of valid tracks > 100. Both CD and Steam versions have 125 tracks, so 100 seems like a safe margin.

### DIFF
--- a/engines/grim/emi/sound/emisound.cpp
+++ b/engines/grim/emi/sound/emisound.cpp
@@ -606,29 +606,15 @@ MusicEntry *EMISound::initMusicTableDemo(const Common::String &filename) {
 	return musicTable;
 }
 
-MusicEntry *EMISound::initMusicTableRetail(MusicEntry *table, const Common::String &filename) {
+void EMISound::initMusicTableRetail(MusicEntry *musicTable, const Common::String filename) {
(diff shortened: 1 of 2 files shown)


engines/grim/emi/sound/emisound.cpp
```diff
@@ -680,16 +666,36 @@ void EMISound::initMusicTable() {
...
+		int numTracks = 0;
+		for (int i = 0; i < 126; i++) {
+			if (!musicTable[i]._filename.empty()) {
+				numTracks++;
+			}
+		}
+
+		warning("Found %d music tracks, expected 125", numTracks);
+		if (numTracks < 100) {
+			delete[] musicTable;
```
Botje: This case does not set `_numMusicStates` or `_musicTable`.

PR #1550, opened 2019-06-01
Update detection.cpp

federicorosso1993:
Added italian version support

### DIFF
--- a/engines/stark/detection.cpp
+++ b/engines/stark/detection.cpp
@@ -177,6 +177,34 @@ static const ADGameDescription gameDescriptions[] = {
 		ADGF_UNSTABLE,
 		GUIO_NONE
 	},
+	
+	// The Longest Journey
+	// Italian DVD version
+	{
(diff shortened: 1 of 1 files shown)


bgK:
Thank you for your contribution.

Are the DVD version checksums from an actual Italian DVD or were they constructed by analogy using the other detection entries / by mixing a DVD English version and a 4 CD Italian version?

federicorosso1993:
Yes there is an official "rare" italian DVD release, here an example:
https://www.ebay.it/itm/254217177224

I asked for both checksum files from people that own the italian version to check if that was the real checksum before change the detection.cpp file (I own both the gog version of the game and the funcom digital version of the game that is not sold anymore but that was available with the italian language):
https://steamcommunity.com/app/6310/discussions/0/626329821115109091/?l=danish

I don't own any physical release

bgK:
Ok, thanks!

PR #1553, opened 2019-06-04
Rework Android build system, provide Travis build scripts.

Botje:
This PR updates the Android build system to fit the modern Gradle approach, which simplifies a lot of stuff and should allow us to get daily Android builds. It also takes a first step towards working on modern Android versions.

The last commit adds a travis.yml that should — at the very least — enable anyone to build ResidualVM and its dependencies.

Comments welcome!
Tentative TODO list:
- [x] Rebase on master
- [ ] Investigate travis `os: windows` together with Conan for dependencies
- [ ] Test Android build and fix SD card access

PR #1556, opened 2019-06-17
BUILD: Write travis file for Ubuntu, OSX, Windows and Android

Botje:
As requested in #1553, this is just the Travis changes for building (and optionally deploying) automatically. It will not do anything by itself, Travis needs to be enabled for the repository.

### DIFF
--- a/.travis/deploy.sh
+++ b/.travis/deploy.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+if [ -z "$DEPLOY_KEY" ] || [ -z "$DEPLOY_HOST" ]; then
+	echo 'Not uploading artifacts; one of DEPLOY_{KEY,HOST} is not set'
+	exit 0
+fi
+
+base64 -d </dev/null >/dev/null 2>&1 && BASE64="base64 -d" || BASE64="base64 -D"
+echo "$DEPLOY_KEY" | $BASE64 > /tmp/upload.key
+chmod 600 /tmp/upload.key
+
+export SSH_ASKPASS=/bin/echo
+echo "put artifacts/* $DEPLOY_PATH" > /tmp/sftp.batch
+sftp -b /tmp/sftp.batch -r -i /tmp/upload.key \
+     -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
+     "$DEPLOY_HOST"
--- a/.travis/freetype-config
+++ b/.travis/freetype-config
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+root=$TRAVIS_BUILD_DIR/dependencies
+
+for flag in "$@"; do
+	case "$flag" in
+		--libs) echo "-L$root/lib -lfreetype -lz" ;;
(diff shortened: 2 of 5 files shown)


.travis/freetype-config
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
aquadran: it's used by android target, maybe rename file to android-freetype-config ?

.travis/deploy.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
aquadran: Is there some documentation about this step?

aquadran: exactly $DEPLOY_KEY, $DEPLOY_HOST and $DEPLOY_PATH

aquadran: This is custom deployment?

.travis/freetype-config
Botje: Good idea.

.travis/deploy.sh
Botje: This script is only executed if a deploy is requested. It scp's artifacts to `$DEPLOY_HOST:$DEPLOY_PATH` using the base64-decoded `$DEPLOY_KEY`. I wrote it to test the output of my jobs.

On the main repo it could be used to copy binaries to the ResidualVM downloads server.
Alternatively, we could store the results in Amazon S3 or Google Cloud buckets, but that is separate from the current infrastructure and not free.

aquadran: I'm interesting at this point only upload Android builds for daily snapshots. But rest use for CI as addition to Buildbot. Or use different folder for Travis builds and upload all. I think last part make more sense at this point.

bgK: > I'm interesting at this point only upload Android builds for daily snapshots.

I personally would prefer all daily builds to come from Travis. The toolchains and build configuration used by Travis are fully open source and make it possible for anybody to make improvements. On the other hand, buildbot has been broken / provided suboptimal builds multiple times in the past because only very few people can make changes to its configuration.

Botje:
Just a small status update: the recent changes to master broke the Windows and Debian builds, while the changes to Travis broke Android and macOS (already fixed).

.travis/deploy.sh
aquadran: Daily or CI builds could be hand off to Travis where possible to save CPU time. Other builds still be here. It need to tuned to have both on the same folder. The builds where not broken on buildbot, but commits broke MacOS and AmigaOS targets. Current Jenkins builds are limited and allow merge broken commits. Linux builds where switched off at some point due host transfer. I wanted to enabled them again, but I forgot it about :( Now started changing Linux targets and moving into Travis at some point. Also I want update AmigaOS target. Windows daily builds are/will be provided manual way as usually. What I don't know yet how to make Travis CI and daily builds. As daily builds are not enough. Also current Travis builds are very limited to targets. Anyway I'm going to merge this PR.

PR #1574, opened 2019-09-07
GRAPHICS: Do not use SDL_WINDOW_RESIZEABLE

raziel-:
See also the forums thread for further information
https://forums.residualvm.org/viewtopic.php?f=1&t=766

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -196,7 +196,8 @@ void OpenGLSdlGraphicsManager::createOrUpdateScreen() {
 	}
 
 	if (!createOrUpdateGLContext(_engineRequestedWidth, _engineRequestedHeight,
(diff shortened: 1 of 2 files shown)


raziel-:
Followup to @Botje 
https://forums.residualvm.org/viewtopic.php?f=1&t=766#p4323

bgK:
Please leave `SDL_WINDOW_RESIZEABLE`, the window is intended to be resizeable.

What needs to be done here is to make the window not resizeable when the game does not support arbitrary resolutions and the platform does not support rendering to a framebuffer.

raziel-:
That is beyond my capabilities, sorry

bgK:
I've pushed a change to your branch to implement that rule. Would you mind testing it on AmigaOS ?

raziel-:
@bkK

With your change it still works on AmigaOS4.
I get a window missing the resize gadget.

bgK:
Ok, let's merge it then.

PR #1578, opened 2019-10-13
MACOS: Fix scaling in Catalina

tsoliman:
Thanks @bgK for the fix


tsoliman:
this is the wrong fix as it prevents HiDPI completely (and also doesn't work without a bundle)
 .. the correct fix to do `[foo setWantsBestResolutionOpenGLSurface:NO]` where foo is the view.
I don't know enough to implement the change.

bgK:
There's been a fix in SDL: https://hg.libsdl.org/SDL/rev/46b094f7d20e

tsoliman:
I had to rebuild libsdl2 locally off of 1.0.10 with just that one patch applied, and that totally fixed it.
Thanks again @bgK !

PR #1582, opened 2019-11-06
MACOS: Generate bundle without external dependencies

Botje:
This PR adds the necessary build commands to bundle libSDL2.dylib with the app. It should allow dependency-less 64-bit OSX apps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,10 +78,9 @@ matrix:
             - glew
             - bzip2
             - zlib
-            - libiconv
             - freetype
       install:
-        - cp -v /usr/local/Cellar/{bzip2,libiconv,zlib}/*/lib/*.a /usr/local/lib
+        - cp -v /usr/local/Cellar/{bzip2,zlib}/*/lib/*.a /usr/local/lib
       script:
         - ./configure --enable-all-engines && make -j4
       before_deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ matrix:
       install:
         - cp -v /usr/local/Cellar/{bzip2,zlib}/*/lib/*.a /usr/local/lib
       script:
-        - ./configure --enable-all-engines && make -j4
+        - ./configure --enable-all-engines && make -j4 bundle
       before_deploy:
         - make osxsnap
         - mkdir artifacts && cp -r ResidualVM-snapshot.dmg artifacts
--- a/ports.mk
+++ b/ports.mk
@@ -177,7 +177,10 @@ endif
 # Location of static libs for the iPhone
 ifneq ($(BACKEND), iphone)
 # Static libaries, used for the residualvm-static and iphone targets
-OSX_STATIC_LIBS := `$(SDLCONFIG) --static-libs`
+OSX_STATIC_LIBS := $(shell $(SDLCONFIG) --static-libs)
+ifdef USE_SDL2
+OSX_STATIC_LIBS := $(subst -lSDL2,$(STATICLIBPATH)/lib/libSDL2.a,$(OSX_STATIC_LIBS))
+endif
 ifdef USE_SDL_NET
 ifdef USE_SDL2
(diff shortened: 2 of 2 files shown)


bgK:
Is there a problem with statically linking SDL2? The binary from macports works fine in my experience, and is compatible with macOS 10.6.

Botje:
Replacing `-lSDL2` with `/usr/local/lib/libSDL2.a` in the `residualvm-static` target gives me:
```
Undefined symbols for architecture x86_64:
  "_iconv", referenced from:
      _SDL_iconv_REAL in libSDL2.a(SDL_iconv.o)
     (maybe you meant: _SDL_iconv_open_REAL, _SDL_iconv , _SDL_iconv_string , _SDL_iconv_open , _SDL_iconv_close_REAL , _SDL_iconv_REAL , _SDL_iconv_string_REAL , _SDL_iconv_close )
  "_iconv_close", referenced from:
      _SDL_iconv_close_REAL in libSDL2.a(SDL_iconv.o)
      _SDL_iconv_string_REAL in libSDL2.a(SDL_iconv.o)
     (maybe you meant: _SDL_iconv_close_REAL, _SDL_iconv_close )
  "_iconv_open", referenced from:
      _SDL_iconv_open_REAL in libSDL2.a(SDL_iconv.o)
      _SDL_iconv_string_REAL in libSDL2.a(SDL_iconv.o)
     (maybe you meant: _SDL_iconv_open_REAL, _SDL_iconv_open )
```
It appears to be picking up libiconv from my brew install instead of from /usr/lib.

Botje:
Do the above changes work for you? It seems brew SDL2 is compiled against system libiconv, so we have to link against that as well. If Macports' SDL2 uses its own libiconv, we have a problem :)

bgK:
> Do the above changes work for you? It seems brew SDL2 is compiled against system libiconv, so we have to link against that as well. If Macports' SDL2 uses its own libiconv, we have a problem :)

Unfortunately they don't:
```Undefined symbols for architecture x86_64:
  "_libiconv", referenced from:
      Common::convertToU32String(Common::Encoding, Common::String const&) in libcommon.a(iconv.o)
  "_libiconv_close", referenced from:
      Common::convertToU32String(Common::Encoding, Common::String const&) in libcommon.a(iconv.o)
  "_libiconv_open", referenced from:
      Common::convertToU32String(Common::Encoding, Common::String const&) in libcommon.a(iconv.o)
```

This is with my osx-cross toolchain. The toolchain environment is:
```
AR="x86_64-apple-darwin12-ar"
OSXCROSS_MP_INC=1
PKG_CONFIG_SYSROOT_DIR="/opt/osxcross/target/macports/pkgs"
MACOSX_DEPLOYMENT_TARGET="10.6"
PATH="/opt/osxcross/target/bin:/opt/osxcross/target/macports/pkgs/opt/local/bin:$PATH"
PKG_CONFIG_LIBDIR="/opt/osxcross/target/macports/pkgs/opt/local/lib/pkgconfig"
```
ResidualVM is configured with:
`../configure --host=x86_64-apple-darwin12 --enable-static --with-staticlib-prefix=/opt/osxcross/target/macports/pkgs/opt/local`

Botje:
Then we will need to check (using `otool`, not sure if that is available in osx-cross?) which libiconv libSDL2 is linked to and make sure to either link against the corresponding static library or bundle the .so.

Alternatively we could re-add the `OSX_STATIC_LIBS += $(STATICLIBPATH)/lib/libiconv.a` bit but make it conditional on this file existing.

Which approach do you see working?

bgK:
> Then we will need to check (using otool, not sure if that is available in osx-cross?) which libiconv libSDL2 is linked to and make sure to either link against the corresponding static library or bundle the .so.

SDL dynamically loads iconv if it is available. I's our own code that links against iconv.

> Which approach do you see working?

IMO the correct approach is to rewrite this whole mess using this PR as a starting point. Hardcoding the libraries is never going to be robust:
https://github.com/scummvm/scummvm/pull/1307/files

aquadran:
There is no issue with static linking, no external libs.

Issue #1519, opened 2019-02-05
EMI: On RPI, missing transitions in introduction and video stucks after it


lmerckx:
Hello,

I'm testing EMI (French version) on a Raspberry PI 2 (but same problem with PI 3).
I'm using a OpenGL ES 2 driver with hardware acceleration. Here are some information for more details:
```
OpenGL maximum texture size: 2048
OpenGL extensions: GL_OES_compressed_ETC1_RGB8_texture GL_OES_compressed_paletted_texture GL_OES_texture_npot GL_OES_depth24 GL_OES_vertex_half_float GL_OES_EGL_image GL_OES_EGL_image_external GL_EXT_discard_framebuffer GL_OES_rgb8_rgba8 GL_OES_depth32 GL_OES_mapbuffer GL_EXT_texture_format_BGRA8888 GL_APPLE_rgb_422 GL_EXT_debug_marker 
OpenGL GLSL version: 100
OpenGL: GLES2 context initialized
OpenGL: NPOT texture support: 1
OpenGL: Shader support: 1
OpenGL: FBO support: 1
OpenGL: Packed depth stencil support: 0
OpenGL: Unpack subimage support: 0
INFO: OpenGL Vendor: Broadcom
INFO: OpenGL Renderer: VideoCore IV HW
INFO: OpenGL Version: OpenGL ES 2.0
INFO: OpenGL Red bits: 8
INFO: OpenGL Green bits: 8
INFO: OpenGL Blue bits: 8
INFO: OpenGL Alpha bits: 8
INFO: OpenGL Z buffer depth bits: 24
INFO: OpenGL Double Buffer: 1
INFO: OpenGL Stencil buffer bits: 8
```

There are some minor glitches during the introduction:
- CD icon missing when skipping introduction
- transitions missing: Lucasarts logo transforms to monkey without fading, turning page animation missing, missing background page when opening menu, ...

But the major problem is that the graphics stuck after the introduction. The sound go on, the dialog propositions are displayed but no image.
Could it be a problem with the GLES library ? Some missing features ?

Thank you in advance.

lmerckx:
Hello,

I'm still trying to solve this problem. So, I added some `glGetError()` after each GL call to get more information.
[...]

PR #1584, opened 2019-11-19
Better support of OpenGLES (ex: Raspberry PI)

lmerckx:
Description:
- Uses GL_UNSIGNED_SHORT for glDrawElements instead of GL_UNSIGNED_INT (not supported for GLES2)
- Change Light structure for GLES2 compatibility: array of struct -> 4 arrays
- Split shaders for billboard, lights enabled and disabled: avoid OutOfMemory on Raspberry

See #1519 for full history of problem and resolution

### DIFF
--- a/engines/grim/emi/modelemi.cpp
+++ b/engines/grim/emi/modelemi.cpp
@@ -36,10 +36,10 @@
 namespace Grim {
 
 struct Vector3int {
-	int _x;
-	int _y;
-	int _z;
-	void setVal(int x, int y, int z) {
+	uint16 _x;
+	uint16 _y;
+	uint16 _z;
+	void setVal(uint16 x, uint16 y, uint16 z) {
 		_x = x; _y = y; _z = z;
 	}
 };
@@ -79,9 +79,9 @@ void EMIMeshFace::loadFace(Common::SeekableReadStream *data) {
 			y = data->readSint32LE();
 			z = data->readSint32LE();
 		} else {
-			x = data->readSint16LE();
-			y = data->readSint16LE();
-			z = data->readSint16LE();
+			x = data->readUint16LE();
+			y = data->readUint16LE();
+			z = data->readUint16LE();
 		}
 		_indexes[j++].setVal(x, y, z);
 	}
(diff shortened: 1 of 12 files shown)


engines/grim/emi/modelemi.cpp
```diff
@@ -79,9 +79,9 @@ void EMIMeshFace::loadFace(Common::SeekableReadStream *data) {
 			y = data->readSint32LE();
 			z = data->readSint32LE();
 		} else {
-			x = data->readSint16LE();
-			y = data->readSint16LE();
-			z = data->readSint16LE();
+			x = data->readUint16LE();
```
klusark: This can't be negative?

engines/grim/emi/modelemi.cpp
```diff
@@ -36,10 +36,10 @@
...
 
 struct Vector3int {
-	int _x;
-	int _y;
-	int _z;
-	void setVal(int x, int y, int z) {
+	uint16 _x;
+	uint16 _y;
+	uint16 _z;
+	void setVal(uint16 x, uint16 y, uint16 z) {
```
klusark: Won't this break the PS2 models?

engines/grim/emi/modelemi.cpp
lmerckx: I assumed no. 

I'm not a specialist of OpenGL but the only types supported for indices in glDrawElements seem GL_UNSIGNED_BYTE, GL_UNSIGNED_SHORT or GL_UNSIGNED_INT (depending on the version).
I wonder if the same modification cannot be done for PS2 version too.

engines/grim/emi/modelemi.cpp
lmerckx: I've just tried the PS2 version and it still works.
At least the introduction, the boat chapter and the beginning on Melee Island.

engines/grim/emi/modelemi.cpp
Botje: `_indexes` are indices into a larger array of vertices, so cannot be negative.

lmerckx: Thank you for you confirmation Botje. So, I can make the same modification for PS2 version ?

Botje: Yep!

lmerckx: Done and committed.

aquadran:
I think it would be better to split each changes here

aquadran:
please port this into scummvm PR, smaller commits or few PRs

aquadran:
PR is now on ScummVM side

Issue #1568, opened 2019-08-05
ANDROID: Need 64-bit builds


Botje:
As of August 1, 2019, Google will [reject](https://developer.android.com/distribute/best-practices/develop/64-bit) applications that do not have 64-bit support.
My plan is to generate these via Travis and upload builds as commits come in.

We need:
- [x] Travis configuration for building 64-bit and 32-bit side by side (I have this locally and will pull-request soonish)
- [ ] Travis configuration for uploading applications to Google Play
- [ ] API access and a service account on the Google Play Developer Console (I think only @klusark can do this)

klusark:
@Botje Is that an account just for travis? Let me know the details and I can set it up.

Botje:
@klusark Yes, it would be a service account just for uploading APKs (and possibly pushing them to a beta channel). I think the instructions are [here](https://developers.google.com/android-publisher/getting_started#using_a_service_account), but I've never done it myself.

Botje:
With the latest pull request merged we are building 64-bit Android apps.
Now to find out how to do the signing...

Botje:
@klusark apparently you got an intermediate signing key from Google when uploading your final signing key. Can you either send it to me or add it to the repo using [these instructions](https://docs.travis-ci.com/user/encrypting-files/) ? The end result should be a file in `.travis/android-signing-key.enc` and a decryption command.

Smashman4:
Has there been any updates on this?

PR #1585, opened 2019-11-25
ANDROID: Generate universal signed APKs

Botje:
This PR addresses the second part of #1568.

It builds universal (ie both 32bit and 64bit ARM) APKs and signs them with the upload key.
This key is stored in an encrypted keystore, that can only be unlocked on the master branch here.
Finally, the version code tracks Travis build numbers so we can (later) automatically push builds to Google Play.


klusark:
👍

PR #1592, opened 2020-01-29
GRIM: Fix memory issues

mparnaudeau:
These changes fix issues with a minor impact because they are revealed at exit.
But that cleans the code and avoid pollution, what will ease the detection of other problems.

Problems were found thanks to compilation code instrumentation (using sanitizers).

### DIFF
--- a/engines/grim/gfx_opengl.cpp
+++ b/engines/grim/gfx_opengl.cpp
@@ -113,6 +113,8 @@ GfxOpenGL::GfxOpenGL() : _smushNumTex(0),
 GfxOpenGL::~GfxOpenGL() {
 	delete[] _storedDisplay;
 
+	releaseMovieFrame();
+
 	if (_emergFont && glIsList(_emergFont))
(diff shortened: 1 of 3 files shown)


mparnaudeau:
@klusark Thank you for your feedback.

About the leak on MovieFrame, did you have a look? As far as I see, I don't know when releaseMovieFrame would be called from GrimEngine::updateDisplayScene and GrimEngine::drawNormalMode. But you're right about the problem in other engines: the call to releaseMovieFrame was already present in the TinyGL case but not in the OpenGL case (I added it) and OpenGL shaders case (I will add and test it).

I will also propose an alternative about the fix around setKey.

That takes time for me to understand the code because that's my first attempt. But fixing this kind of bug really suits this discovery.

mparnaudeau:
Trying to check some behaviours regarding problems related here, with different graphics engines, I faced other problems so I discard this pull request and will create more specific ones.

PR #1605, opened 2020-02-23
BUILD: Apply LDFLAGS on modules/plugins

besser82:
Some distributions populate their build environment with special linker flags, e.g. for hardening purposes such as immediate binding and/or read-only relocations.

Those flags should be honored when linking dynamic libraries.

This approach uses the unmodified LDFLAGS, stored in the SAVED_LDFLAGS variable by the configure script, from the system environment for linking dynamic plugins, which is very likely not to cause any harm.

Backported from scummvm/scummvm@0974f168.

See https://github.com/scummvm/scummvm/pull/2082 for further reference.  https://github.com/scummvm/scummvm/pull/2077 contains some links to documentation what binary hardening is.


besser82:
@sev- Can you take care of this, please?

aquadran:
It will merge along with sync with ScummVM

besser82:
> It will merge along with sync with ScummVM

@aquadran Can you please make sure this will be merged into https://github.com/residualvm/residualvm-tools, too?

Issue #1464, opened 2018-06-28
STARK: Inconsistency in debug commands printing "too few args"


Faalagorn:
Another #1456-related change.

It's only a minor visual issue, but some commands currently print the "Too few args" information when are invoked without any argument, and some do not. I'm not sure how other engines (including ScummVM ones) implement this, but based on the built-in commands (`debugLevel`, `debugflag_enable` and `debugflag_disable` typing the commands without any argument should only print the usage instructions and not the warning about the too few arguments (unless it's of course a command passed with less argument that actually is needed, for example `changeKnowledge`)

The commands that currently print the "Too few args" message when invoked without any argument:
* `changeKnowlede`
* `decompileScript`
* `enableScript`
* `forceScript`

Commands that **do not** print the "Too few args" message when invoked without any arguments include:
* `changeChapter`
* `changeLocation`
* `dumpArchive`
* `enableInventoryItem`

Other commands are invoked without any arguments normally.

PR #1609, opened 2020-02-28
STARK: Make debug command output more consistent

GunnarBirke:
The 'Too few args' message should only be
displayed if some but not enough arguments
are provided. Otherwise just show the
instructions of the command.

Refs #1464


bgK:
Good, thanks!

PR #1619, opened 2020-04-08
MYST3: Fix default text language when "text_language" key doesn't exist in residualvm.ini

usineur:
Currently, "text_language" key is not set just after adding "Myst 3" in residualvm.

It causes text to be set to "english" by default. This commit fixes that.


bgK:
Thanks, I have pushed a slightly different version of your fix.

PR #1640, opened 2020-05-17
EMI: Always enable movie subtitles for the demo

### LINKED ISSUE #1161: EMI: Subtitles are missing in EMI demo-intro (French/German)
Atleast the French version of the EMI demo has english voice, but german subs in the intro video, these are missing in ResidualVM.

Botje:
This fixes #1161.

In the course of implementing this, I stumbled upon a race condition in the subtitle drawing:

- the movie thread takes the frame mutex to update its internal structures and update `_movieSubtitle`.
- the drawing thread only takes the mutex (briefly) to grab any new movie frames, but does not synchronize to take the subtitle.

I fixed this by drawing the subtitles while holding the movie mutex. It works but I'm open to nicer suggestions. I tried to see if we could only take a _copy_ of the `TextObject::Ptr` while holding the mutex, but the movie player calls `delete` on the underlying pointer, and a `TextObject::Ptr` is not a proper smart pointer...

### DIFF
--- a/engines/grim/emi/lua_v2.cpp
+++ b/engines/grim/emi/lua_v2.cpp
@@ -171,6 +171,9 @@ void Lua_V2::StartMovie() {
 			showSubtitles = true;
 		}
 	}
+	if (g_grim->getGameFlags() & ADGF_DEMO) {
+		showSubtitles = true;
+	}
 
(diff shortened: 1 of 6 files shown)


aquadran:
👍

PR #1642, opened 2020-05-27
AMIGAOS4: Automate installation process on local builds

raziel-:
There´s a catch unfortunately.

As one can, see there are two files automatically installed (icons for drawer and app), but i´m not allowed to upload those through the GitHub site.

So, i´d need help with that, please.
Could some dev please grab the attached files and place them into "dists/amiga/"?

Thanks a lot
[residualvm.zip](https://github.com/residualvm/residualvm/files/4688502/residualvm.zip)


raziel-:
Thank you very much

PR #1651, opened 2020-06-15
GSOC WME3D: Opengl3d renderer 2d additions

GunnarBirke:


engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -128,8 +128,27 @@ bool BaseSurfaceOpenGL3D::create(const Common::String &filename, bool defaultCK,
...
+				_surf->format.colorToARGB(pix, a, r, g, b);
+				if (r == ckRed && g == ckGreen && b == ckBlue) {
+					a = 0;
+					r = 0;
+					g = 0;
+					b = 0;
+					((uint32 *)_surf->getPixels())[i * _surf->w + j] = _surf->format.ARGBToColor(a, r, g, b);
+				} else if (replaceAlpha) {
+					a = 255;
+					((uint32 *) _surf->getPixels())[i * _surf->w + j] = _surf->format.ARGBToColor(a, r, g, b);
```
aquadran, line 150: remove space from "((uint32 *) _surf->getPixels())"

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.h
```diff
@@ -69,6 +70,7 @@ class BaseSurfaceOpenGL3D : public BaseSurface {
 	void setTexture();
 
 private:
+	Graphics::Surface *_surf;
 	OpenGL::Texture* tex;
```
somaen: These should have a underscore-prefix.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -128,8 +128,27 @@ bool BaseSurfaceOpenGL3D::create(const Common::String &filename, bool defaultCK,
 	}
 
 	if (needsColorKey) {
-		Graphics::TransparentSurface trans(*_surf);
-		trans.applyColorKey(_ckRed, _ckGreen, _ckBlue, replaceAlpha);
+		// this is taken from Graphics::TransparentSurface
+		// only difference is that we set the pixel
+		// color to transparent black, like D3DX,
+		// if it matches the color key
+		for (int i = 0; i < _surf->h; i++) {
```
somaen, line 139: May I suggest using x and y as iteration variables (as appropriate), instead of i and j. More meaningful, and easier to verify that they make sense then

engines/wintermute/video/video_theora_player.cpp
```diff
@@ -143,7 +143,11 @@ bool VideoTheoraPlayer::initialize(const Common::String &filename, const Common:
 
 	// Additional setup.
 	_surface.create(_theoraDecoder->getWidth(), _theoraDecoder->getHeight(), _theoraDecoder->getPixelFormat());
+#ifdef ENABLE_WME3D
+	_texture = _gameRef->_renderer->createSurface();
```
somaen, line 147: I wonder if there's any reason why the original code isn't using this function, because I get the feeling that is should be doing that. In which case we don't need to special-case for WME_3D, as the renderer should be handling this detail.

engines/wintermute/ad/ad_actor_3dx.cpp
```diff
@@ -166,13 +166,13 @@ bool AdActor3DX::update() {
 			float walkVel = _directWalkVelocity == 0.0f ? _velocity : _directWalkVelocity;
 			Math::Vector3d newPos = _posVector;
 			if (_directWalkMode == DIRECT_WALK_FW) {
-				newPos.x() += -sinf(_angle.getRadians()) * walkVel * _scale3D * (float)_gameRef->_deltaTime / 1000.f;
```
somaen, line 169: I would suggest making some short comments around a lot of these changes, to document the LH -> RH conversion. In case someone at some later point tries to debug some of this, and checks against original WME, and then get confused.

engines/wintermute/base/base_object.cpp
```diff
@@ -102,7 +102,7 @@ BaseObject::BaseObject(BaseGame *inGame) : BaseScriptHolder(inGame) {
 	_shadowModel = nullptr;
 	_posVector = Math::Vector3d(0.0f, 0.0f, 0.0f);
 	_angle = 0.0f;
-	_scale = 1.0f;
+	_scale3D = 1.0f;
```
somaen, line 105: Consider running something like the static-analyzer in Clang to see if there's other similar mistakes.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -188,6 +179,21 @@ bool BaseSurfaceOpenGL3D::create(int width, int height) {
 	return true;
 }
 
+bool BaseSurfaceOpenGL3D::putSurface(const Graphics::Surface &surface, bool hasAlpha) {
```
somaen, line 182: hasAlpha is unused.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.h
```diff
@@ -50,9 +50,7 @@ class BaseSurfaceOpenGL3D : public BaseSurface {
 	bool restore() override;
 	bool create(const Common::String &filename, bool defaultCK, byte ckRed, byte ckGreen, byte ckBlue, int lifeTime = -1, bool keepLoaded = false) override;
 	bool create(int width, int height) override;
-	bool putSurface(const Graphics::Surface &surface, bool hasAlpha = false) override {
-		return STATUS_FAILED;
-	}
+	bool putSurface(const Graphics::Surface &surface, bool hasAlpha = false);
```
somaen, line 53: I assume this should still have override?

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
```diff
@@ -20,6 +20,7 @@
  *
  */
 
+#include "backends/platform/sdl/sdl-sys.h"
```
somaen, line 23: Including the backend is not allowed, as we may have non-SDL backends. There is some limited use of such includes in the grim-engine, but those are themselves guarded with #if(SDL_BACKEND)

What's wrong with the values passed to initRenderer?

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -128,8 +128,27 @@ bool BaseSurfaceOpenGL3D::create(const Common::String &filename, bool defaultCK,
 	}
 
 	if (needsColorKey) {
-		Graphics::TransparentSurface trans(*_surf);
-		trans.applyColorKey(_ckRed, _ckGreen, _ckBlue, replaceAlpha);
+		// this is taken from Graphics::TransparentSurface
```
somaen, line 135: Consider putting this functionality in a separate function.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -48,10 +48,17 @@ bool BaseSurfaceOpenGL3D::displayHalfTrans(int x, int y, Rect32 rect) {
 }
 
 bool BaseSurfaceOpenGL3D::isTransparentAt(int x, int y) {
-	// wme3d checks for a <= 128 in this function
-	// and for a == 0 in isTransparentAtLite
-	// for the moment, let's do it like the software renderer
-	return isTransparentAtLite(x, y);
+	uint8 *imageData = new uint8[4 *  _width * _height];
```
somaen, line 51: Add () to get zero-initialization.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -48,10 +48,17 @@ bool BaseSurfaceOpenGL3D::displayHalfTrans(int x, int y, Rect32 rect) {
...
+
+	// assume 32 bit rgba for now
+	glBindTexture(GL_TEXTURE_2D, _tex);
+	glGetTexImage(GL_TEXTURE_2D, 0, GL_RGBA, GL_UNSIGNED_BYTE, imageData);
+	glBindTexture(GL_TEXTURE_2D, 0);
+
+	uint8 alpha = *(imageData + y * _width * 4 + x * 4 + 3);
```
somaen, line 58: Use []-notation instead.

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
```diff
@@ -442,7 +442,7 @@ bool BaseRenderOpenGL3D::drawSpriteEx(BaseSurfaceOpenGL3D &tex, const Wintermute
 	vertices[2 * vertexSize + 8 + 2] = b;
 	vertices[2 * vertexSize + 8 + 3] = a;
 	vertices[3 * vertexSize + 8 + 0] = r;
-	vertices[3 * vertexSize + 8 + 2] = g;
+	vertices[3 * vertexSize + 8 + 1] = g;
```
somaen, line 448: I'd suggest using a proper struct instead, or at the very least constants, to reduce the risk of such mistakes.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.h
GunnarBirke: Added override back

engines/wintermute/ad/ad_actor_3dx.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Not sure about how to use it here or how this could be useful

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
GunnarBirke: Actually I had problems with the window size since the max resolution of my laptop is too small in the height dimension for Alpha Polaris. The code above fixed an issue with the mouse cursor, which first would not reach the bottom of the window, probably because the parameters above do not represent the size of the actual drawable area of the window but instead the resolution of the game (which seems fixed for wme games, although wme also seems to try hard to find an almost fitting resolution in case the preferred one is not available). However, I have just plugged in a monitor with a bigger resolution, so for the moment everything is fine anyways. I will remove the code from above.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
GunnarBirke: Put in a struct for this

engines/wintermute/video/video_theora_player.cpp
GunnarBirke: I couldn't find anything in the player which specifically uses something from BaseSurfaceOSystem and it seems to work with the OpenGL surface anyways. Remove the macro guard?

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -386,7 +394,7 @@ bool BaseFileManager::hasFile(const Common::String &filename) {
 	if (diskFileExists(filename)) {
 		return true;
 	}
-	if (_packages.hasFile(filename)) {
+	if (_packages.hasFile(backwardSlashesPath)) {
 		return true;    // We don't bother checking if the file can actually be opened, something bigger is wrong if that is the case.
 	}
 	if (!_detectionMode && _resources->hasFile(filename)) {
```
GunnarBirke, line 400: This fixes the issue about the paths coming from the scripts. If I understand correctly, only reversing slashes for file paths in packages is consistent with the rest of the code. Do we need any macro guards here?

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
```diff
@@ -361,6 +362,18 @@ bool BaseRenderOpenGL3D::drawSprite(BaseSurfaceOpenGL3D &tex, const Wintermute::
 	return drawSpriteEx(tex, rect, pos, Vector2(0.0f, 0.0f), scale, 0.0f, color, alphaDisable, blendMode, mirrorX, mirrorY);
 }
 
+struct SpriteVertex {
```
somaen, line 365: Since the usage of the struct depends on the structs byte-layout, it should be declared with the PACKED_STRUCT attribute:

struct X {
...
} PACKED_STRUCT;

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -372,6 +372,14 @@ uint32 BaseFileManager::getPackageVersion(const Common::String &filename) {
 
 //////////////////////////////////////////////////////////////////////////
 bool BaseFileManager::hasFile(const Common::String &filename) {
+	Common::String backwardSlashesPath = filename;
+	// correct slashes
+	for (uint32 i = 0; i < backwardSlashesPath.size(); i++) {
+		if (backwardSlashesPath[(int32)i] == '/') {
+			backwardSlashesPath.setChar('\\', (uint32)i);
```
somaen, line 379: Unnecessary cast.

engines/wintermute/base/base_file_manager.cpp
somaen: Alpha Polaris might be the first game that happens to combine such queries with backward-slashed filenames. Which would explain why this hasn't been seen before.

BUT:

That doesn't guarantee that we don't have games somehow expecting forward-slashes. 

BUT: As far as I understand we've been unconditionally doing this when opening files, so it should (as far as I know, be safe to do without macro guards.)

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
```diff
@@ -387,58 +421,63 @@ bool BaseRenderOpenGL3D::drawSpriteEx(const OpenGL::Texture &tex, const Wintermu
...
-	vertexCoords[1 * 6 + 3] = pos.x - 0.5f;
-	vertexCoords[1 * 6 + 4] = corrected_y - height - 0.5f;
-	vertexCoords[1 * 6 + 5] = -1.1f;
-	vertexCoords[2 * 6 + 3] = pos.x + width - 0.5f;
-	vertexCoords[2 * 6 + 4] = corrected_y - 0.5f;
-	vertexCoords[2 * 6 + 5] = -1.1f;
-	vertexCoords[3 * 6 + 3] = pos.x + width - 0.5f;
-	vertexCoords[3 * 6 + 4] = corrected_y - height - 0.5;
-	vertexCoords[3 * 6 + 5] = -1.1f;
+	vertices[0].x = pos.x - 0.5f;
```
somaen, line 440: Might as well un-unroll this loop. Less error-prone that way. The compiler may or may not unroll it as it sees fit.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -27,8 +27,49 @@
 
 namespace Wintermute {
 
+template<class T>
```
somaen, line 30: Instead of copying this from the Grim-engine, I'd suggest moving it into math/, and use that version in both places. Finding a power-of-two suitable for texture dimensions is a problem common to most of our engines' GL renderers anyhow.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
somaen: Fair enough.

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -167,36 +242,34 @@ bool BaseSurfaceOpenGL3D::comparePixel(int x, int y, byte r, byte g, byte b, int
 }
 
 bool BaseSurfaceOpenGL3D::startPixelOp() {
-	glBindTexture(GL_TEXTURE_2D, tex->getTextureName());
+	_imageData = new uint8[4 * _texWidth * _texHeight];
```
somaen, line 245: Add trailing () for zero-init

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -167,36 +242,34 @@ bool BaseSurfaceOpenGL3D::comparePixel(int x, int y, byte r, byte g, byte b, int
...
 
-	uint8 *image_data = nullptr;
-
-	// assume 32 bit rgba for now
-	glGetTexImage(GL_TEXTURE_2D, 0, GL_RGBA, GL_UNSIGNED_BYTE, image_data);
-
-	uint32 pixel = *reinterpret_cast<uint32 *>(image_data + y * tex->getWidth() * 4 + x * 4);
-	pixel &= 0x000000FF;
-	return pixel == 0;
+	uint8 alpha = _imageData[y * _texWidth * 4 + x * 4 + 3];
```
somaen, line 267: This assumes the format has a byte order that puts alpha in that position. It would be preferable to query that, as it may not match on Big-Endian systems.

engines/wintermute/video/video_theora_player.cpp
somaen: Well, the type of _texture is BaseSurface*, so any subtype details are lost immediately anyhow. The major point of having that factory create-function, was to avoid having to have knowledge of specific renderers outside of the renderer itself.

So I'd say drop the macro-guard, and replace the old code with your call to createSurface(). (BaseRenderOSystem::createSurface performs the exact same "return new BaseSurfaceOSystem(_gameRef);" anyhow.)

While you're at it, you can modify the #include to include "base_surface.h", instead of "base_surface_osystem.h"

engines/wintermute/base/base_file_manager.cpp
GunnarBirke: This was copied from the existing ScummVM code in openPkgFile, meaning the loop is duplicated now. I didn't introduce a new function to leave the code as is as much as possible. Cast is only removed in hasFile now. Still make an extra function out of it?

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Actually there is a nextHigher2 function in algorithm.h. I will use that one

engines/wintermute/base/base_file_manager.cpp
GunnarBirke: Nevermind, algorithm.h has replace

engines/wintermute/base/gfx/opengl/base_render_opengl3d.cpp
GunnarBirke: Did this for the color values but I don't see how to put this into a loop without putting width and height into another array and doing some index calculations

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: If I am not mistaken, than GL_RGBA + UNSIGNED_BYTE means that the alpha component will have offset three here, at least if the texture was uploaded correctly. On the other hand, there are certainly more places where endianess might become an issue. I will have to test things on a big endian platform at some point, I guess.

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -34,6 +34,7 @@
 #include "engines/wintermute/base/file/base_package.h"
 #include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/wintermute.h"
+#include "common/algorithm.h"
```
aquadran, line 37: missing ENABLE_WME3D

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -372,6 +373,10 @@ uint32 BaseFileManager::getPackageVersion(const Common::String &filename) {
 
 //////////////////////////////////////////////////////////////////////////
 bool BaseFileManager::hasFile(const Common::String &filename) {
+	Common::String backwardSlashesPath = filename;
```
aquadran, line 376: missing ENABLE_WME3D

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -386,7 +391,7 @@ bool BaseFileManager::hasFile(const Common::String &filename) {
 	if (diskFileExists(filename)) {
 		return true;
 	}
-	if (_packages.hasFile(filename)) {
+	if (_packages.hasFile(backwardSlashesPath)) {
```
aquadran, line 394: missing ENABLE_WME3D

engines/wintermute/base/base_object.cpp
aquadran, line 105: missing ENABLE_WME3D

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
aquadran: add some comment here about potential endian issue

engines/wintermute/video/video_theora_player.cpp
```diff
@@ -30,7 +30,7 @@
 #include "engines/wintermute/video/video_theora_player.h"
 #include "engines/wintermute/base/base_game.h"
 #include "engines/wintermute/base/base_file_manager.h"
-#include "engines/wintermute/base/gfx/osystem/base_surface_osystem.h"
+#include "engines/wintermute/base/gfx/base_surface.h"
```
aquadran, line 33: add ENABLE_WME3D

engines/wintermute/video/video_theora_player.cpp
```diff
@@ -143,7 +143,7 @@ bool VideoTheoraPlayer::initialize(const Common::String &filename, const Common:
 
 	// Additional setup.
 	_surface.create(_theoraDecoder->getWidth(), _theoraDecoder->getHeight(), _theoraDecoder->getPixelFormat());
-	_texture = new BaseSurfaceOSystem(_gameRef);
+	_texture = _gameRef->_renderer->createSurface();
```
aquadran, line 146: missing ENABLE_WME3D

engines/wintermute/base/base_object.cpp
GunnarBirke: Actually this is already contained in an #ifdef

engines/wintermute/base/base_file_manager.cpp
GunnarBirke: Done

engines/wintermute/video/video_theora_player.cpp
GunnarBirke: Done

engines/wintermute/video/video_theora_player.cpp
GunnarBirke: Done

engines/wintermute/base/base_file_manager.cpp
GunnarBirke: Done

engines/wintermute/base/base_file_manager.cpp
GunnarBirke: Done

engines/wintermute/ad/ad_scene_geometry.cpp
```diff
@@ -672,33 +672,33 @@ bool AdSceneGeometry::directPathExists(Math::Vector3d *p1, Math::Vector3d *p2) {
...
 
-	//			D3DXPLANE plane;
-	//			Math::Vector3d intersection;
-	//			float dist;
+			if (lineSegmentIntersectsTriangle(*p1, *p2, v0, v1, v2, intersection, dist)) {
+				if (lineIntersectsTriangle(*p1, *p1 - *p2, v0, v1, v2, intersection.x(), intersection.y(), intersection.z())) {
+					return intersection;
+				}
+
+				if(lineIntersectsTriangle(*p2, *p2 - *p1, v0, v1, v2, intersection.x(), intersection.y(), intersection.z())) {
```
aquadran, line 695: missing space after if

GunnarBirke: Space inserted

engines/wintermute/base/base_file_manager.cpp
somaen: I'd actually suggest not having that macro-guard here, the fix should be applicable to all games (and makes the behaviour consistent with the file loading).

engines/wintermute/base/base_file_manager.cpp
somaen: Same here.

engines/wintermute/video/video_theora_player.cpp
somaen: Disagree, instead drop both macro guards in this file, so that the only call is to createSurface.

That's how it should have worked anyhow, and 100% equivalent for the non-3D code.

Downside being that it does introduce a minor merge conflict, so I'd suggest creating a separate PR against ScummVM to get this fixed (along with the hasFile-fix)

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -167,36 +231,35 @@ bool BaseSurfaceOpenGL3D::comparePixel(int x, int y, byte r, byte g, byte b, int
 }
 
 bool BaseSurfaceOpenGL3D::startPixelOp() {
-	glBindTexture(GL_TEXTURE_2D, tex->getTextureName());
+	_imageData = new uint8[4 * _texWidth * _texHeight]();
```
somaen, line 234: Potential leak-opportunity, if startPixelOp is called more than once, without calling endPixelOp inbetween. Perhaps we should add a check that _imageData is nullptr here, and error out in that case?

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
```diff
@@ -20,15 +20,45 @@
...
+	}
+}
+
 BaseSurfaceOpenGL3D::BaseSurfaceOpenGL3D(BaseGame *game, BaseRenderOpenGL3D *renderer)
-    : BaseSurface(game), tex(nullptr), renderer(renderer), pixelOpReady(false) {
+	: BaseSurface(game), _tex(0), _renderer(renderer), _imageData(nullptr), _texWidth(0), _texHeight(0) {
+	glGenTextures(1, &_tex);
+}
+
+BaseSurfaceOpenGL3D::~BaseSurfaceOpenGL3D() {
```
somaen, line 60: Just to be safe, this should also delete[] _imageData;

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.h
```diff
@@ -49,29 +50,26 @@ class BaseSurfaceOpenGL3D : public BaseSurface {
...
+	GLuint getTextureName() {
+		return _tex;
+	}
+
 private:
-	OpenGL::Texture* tex;
-	BaseRenderOpenGL3D* renderer;
-	bool pixelOpReady;
+	GLuint _tex;
+	BaseRenderOpenGL3D* _renderer;
```
somaen, line 69: Asterisk should be aligned to "_renderer";

GunnarBirke: Alignment corrected

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: Done

engines/wintermute/base/gfx/opengl/base_surface_opengl3d.cpp
GunnarBirke: In case _imageData is not the nullptr (which means startPixelOp was already called before), startPixelOp just returns success now. That should make it leak safe (together with the deletion inside the destructor).

somaen: It would make it leak-safe, but potentially also hide what I assume would be an error, perhaps it would be better to at least warn in that case, or perhaps assert?

GunnarBirke: Sorry, I misunderstood your first comment. I put in an assert now, much more bulletproof I guess.

PR #1655, opened 2020-06-28
GSOC WME3D: Savegames

GunnarBirke:


engines/wintermute/system/sys_class_registry.h
```diff
@@ -61,6 +61,9 @@ class SystemClassRegistry {
 	void unregisterClasses();
 public:
 	void registerClasses(); // persistent.cpp
+#ifdef ENABLE_WME3D
+	void register3DClasses(); // persisten.cpp
```
somaen, line 65: persistent.cpp

engines/wintermute/ad/ad_scene_geometry.cpp
```diff
@@ -182,6 +182,8 @@ bool AdSceneGeometry::loadFile(const char *filename) {
 	filenameTmp.replace(filenameTmp.size() - 3, 3, "geometry", 0, 8);
 	AdGeomExt *geomExt = getGeometryExtension(filenameTmp.begin());
 
+	SystemClassRegistry::getInstance()->_disabled = true;
```
somaen, line 185: Why do we need to do this? A comment would be good.

engines/wintermute/base/base_persistence_manager.cpp
```diff
@@ -887,12 +887,16 @@ bool BasePersistenceManager::transferMatrix4(const char *name, Math::Matrix4 *va
 bool BasePersistenceManager::transferAngle(const char *name, Math::Angle *val) {
 	if (_saving) {
 		putFloat(val->getDegrees());
+
```
aquadran, line 890: missing #ifdef ENABLE_WME3D

engines/wintermute/base/base_persistence_manager.cpp
```diff
@@ -887,12 +887,16 @@ bool BasePersistenceManager::transferMatrix4(const char *name, Math::Matrix4 *va
...
+		if (_saveStream->err()) {
+			return STATUS_FAILED;
+		}
 	} else {
 		*val = getFloat();
-	}
 
-	if (_loadStream->err()) {
-		return STATUS_FAILED;
+		if (_loadStream->err()) {
```
aquadran, line 897: #ifdef ENABLE_WME3D

engines/wintermute/base/base_object.cpp
```diff
@@ -1032,6 +1037,21 @@ bool BaseObject::persist(BasePersistenceManager *persistMgr) {
 	persistMgr->transferFloat(TMEMBER(_sFXParam3));
 	persistMgr->transferFloat(TMEMBER(_sFXParam4));
 
+#ifdef ENABLE_WME3D
+	if (BaseEngine::instance().getFlags() & GF_3D) {
```
somaen, line 1041: This has the downside that these variables will all be uninitialized after loading a game now. Perhaps outfactor the initialization part of the constructor to a new function that gets called both from the constructor and here, to make things slightly safer. Or, we could consider moving the 3d-related variables to a struct of their own that we instantiate in BaseObject, to keep things slightly easier to keep track of.

The reason this _may_ be dangerous, is that I assume that not ALL places that use these variables are guarded by the flag-check above, and thus they may need to have consistent values even in 2D games, just for safety.

engines/wintermute/ad/ad_object_3d.cpp
```diff
@@ -629,18 +629,14 @@ bool AdObject3D::getBonePosition3D(const char *boneName, Math::Vector3d *pos, Ma
 		return false;
 	}
 
-	if (!offset) {
-		*offset = Math::Vector3d(0, 0, 0);
-	}
-
 	Math::Matrix4 bonePosMat = *boneMat * _worldMatrix;
+	*pos = Math::Vector3d(0.0f, 0.0f, 0.0f);
```
somaen, line 624: This assumes that pos is not a nullptr, if pos can't be a nullptr, then it should be a reference, not a pointer, to make it impossible to pass nullptr.

PR #1661, opened 2020-07-30
GSOC WME: Don't render models on title screen

GunnarBirke:


engines/wintermute/ad/ad_scene.cpp
```diff
@@ -1364,13 +1364,18 @@ bool AdScene::displayRegionContent(AdRegion *region, bool display3DOnly) {
...
 		}
+#else
+		if ((_gameRef->_editorMode || !obj->_editorOnly) && (!objects[i]->_is3D || _sceneGeometry)) {
+			obj->display();
+		}
+#endif
 		obj->_drawn = true;
 	}
 
-
```
aquadran, line 1373: remove line, create difference with scummvm

GunnarBirke: Line restored.

PR #1662, opened 2020-08-10
GSOC WME3D: Binary .X lexer

GunnarBirke:


engines/wintermute/base/gfx/x/modelx.cpp
```diff
@@ -104,9 +104,11 @@ bool ModelX::loadFromFile(const Common::String &filename, ModelX *parentModel) {
 	uint32 fileSize = 0;
 	byte *buffer = BaseFileManager::getEngineInstance()->getEngineInstance()->readWholeFile(filename, &fileSize);
 
-	byte *dataFormatBlock = buffer + 8;
+	byte dataFormatBlock[4];
+	Common::copy(buffer + 8, buffer + 11, dataFormatBlock);
```
aquadran, line 108: add some comments about hardcoded values here. like what it shift value is

engines/wintermute/base/gfx/opengl/meshx.cpp
```diff
@@ -517,20 +520,20 @@ bool MeshX::parseTextureCoords(XFileLexer &lexer) {
 
 bool MeshX::parseNormalCoords(XFileLexer &lexer) {
 	// should be the same as _vertex count
-	uint vertexNormalCount = readInt(lexer);
-	assert(vertexNormalCount == _vertexCount);
+	uint vertexNormalCount = lexer.readInt();
+//	assert(vertexNormalCount == _vertexCount);
```
aquadran, line 524: why assert is disabled?

GunnarBirke: The demo of "On the track of dinosaurs" has a model with fewer normals than vertices. So this condition is actually not guaranteed to be true.

engines/wintermute/base/gfx/x/modelx.cpp
GunnarBirke: Added a comment

engines/wintermute/base/gfx/x/modelx.cpp
```diff
@@ -104,9 +104,13 @@ bool ModelX::loadFromFile(const Common::String &filename, ModelX *parentModel) {
 	uint32 fileSize = 0;
 	byte *buffer = BaseFileManager::getEngineInstance()->getEngineInstance()->readWholeFile(filename, &fileSize);
 
-	byte *dataFormatBlock = buffer + 8;
+	// the header of an .X file consists of 16 bytes
+	// bytes 9 to 12 contain a string which can be 'txt', 'bin', 'bzip, 'tzip', depending on the format
+	byte dataFormatBlock[4];
+	Common::copy(buffer + 8, buffer + 11, dataFormatBlock);
+	dataFormatBlock[3] = '\0';
```
Botje, line 111: This will truncate "bzip" to "bzi\0", so the case on line 115 can never fire.

GunnarBirke: Actually I just increased the buffer size now, didn't think about strncmp.

PR #1715, opened 2020-10-08
ALL: Add error checking to PEResources::getResource

vmedea:
For some reason the `game.exe` that comes with the steam version of TLJ is truncated. It is only 95744 bytes and contains a resource table but not the resource data.

This currently results in random data from the binary to be passed into the BMP parser in `DialogBox::loadBackground()`, resulting in a warning `Only Windows v3 & v4 bitmaps are supported`. It could be worse, but it sets developers on the wrong trail toward trying to fix BMP parsing.

Avoid doing this by adding a check that both the seek and the read succeed, which if so, makes loading the resource simply fail.

### DIFF
--- a/common/winexe_pe.cpp
+++ b/common/winexe_pe.cpp
@@ -212,8 +212,17 @@ SeekableReadStream *PEResources::getResource(const WinResourceID &type, const Wi
 		return nullptr;
 
 	const Resource &resource = _resources[type][id][langList[0]];
-	_exe->seek(resource.offset);
-	return _exe->readStream(resource.size);
+	if (!_exe->seek(resource.offset)) {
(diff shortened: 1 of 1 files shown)


vmedea:
Oh the rabbit hole goes deeper. 
```
game.exe.original (size 590691): MS-DOS executable, NE for MS Windows 3.x (EXE)
game.exe.steam (size 95744):     PE32 executable (GUI) Intel 80386, for MS Windows, PECompact2 compressed
```
The steam version is compressed. I didn't know that was a thing. Maybe with an uncompressor it's possible to get at the resource. In any case, `wrestool` was unable to handle it too.

aquadran:
please submit this to scummvm repo

